PR #5, opened 2021-02-02
Add initial verifier implementation

setrofim:
Add an initial implementation of the verifier component along with
fetcher interfaces for endorsements and trust anchor stores.


common/endorsement_plugin.go
```diff
@@ -0,0 +1,169 @@
...
+
+	qresult, err := s.Impl.RunQuery(qd.Name, qd.Args)
+	if err != nil {
+		return err
+	}
+
+	*resp, err = json.Marshal(qresult)
+	return err
+}
+func (s *EndorsementStoreServer) GetSupportedQueries(args interface{}, resp *[]string) error {
```
thomas-fossati, line 71:
```suggestion

func (s *EndorsementStoreServer) GetSupportedQueries(args interface{}, resp *[]string) error {
```

common/plugin.go
```diff
@@ -0,0 +1,104 @@
...
+			rpcClient, err := client.Client()
+			if err != nil {
+				log.Printf("Failed to load RPC client from %v: %v\n", pluginPath, err)
+				client.Kill()
+				continue
+			}
+
+			raw, err := rpcClient.Dispense(plugType)
+			if err != nil {
+				log.Printf("Plugin %v does not implement a %v.\n", pluginPath, plugType)
```
thomas-fossati, line 82:
```suggestion
				log.Printf("Plugin %v does not implement a %v\n", pluginPath, plugType)
```

common/plugin.go
```diff
@@ -0,0 +1,104 @@
...
+				client.Kill()
+				continue
+			}
+
+			named := raw.(INamed)
+			if named.GetName() != plugName {
+				log.Printf("Wrong name in %v.\n", pluginPath)
```
thomas-fossati, line 89:
```suggestion
				log.Printf("Wrong name in %v\n", pluginPath)
```

common/tokenformat.go
```diff
@@ -0,0 +1,53 @@
...
+		return "dice"
+	default:
+		return fmt.Sprintf("TokenFormat(%v)", tf)
+	}
+}
+
+const (
+	// PsaIatToken is the PSA initial attestation token (based on Entity
+	// Attestation Token RFC:
+	// https://datatracker.ietf.org/doc/draft-ietf-rats-eat/)
```
thomas-fossati, line 31:
```suggestion
	// PsaIatToken is the PSA initial attestation token (based on:
	// https://datatracker.ietf.org/doc/draft-tschofenig-rats-psa-token/)
```

plugins/sqliteendorsements/main.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		result = append(result, hwId)
+	}
+
+	return result, nil
+}
+
+func (e *SqliteEndorsementStore) GetSoftwareComponents(args common.QueryArgs) (common.QueryResult, error) {
+	//TODO: this method is horrible; do not read this; it needs to be destroyed with fire
```
thomas-fossati, line 99: :-)

thomas-fossati (approved):
LGTM! I've left a couple of very minor and non blocking suggestions.

common/evidenceextractor_plugin.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (e EvidenceExtractorRPC) ExtractEvidence(token []byte, ta []byte) (map[string]interface{}, error) {
+	args := &ExtractEvidenceArgs{
+		Token:       token,
+		TrustAnchor: ta,
+	}
+
+	var resp []byte
+
```
yogeshbdeshpande, line 111: Is the blank line intentional here ?

common/evidenceextractor_plugin.go
```diff
@@ -0,0 +1,121 @@
...
+	return &EvidenceExtractorRPC{client: c}, nil
+}
+
+type EvidenceExtractorServer struct {
+	Impl IEvidenceExtractor
+}
+
+func (s EvidenceExtractorServer) GetName(args interface{}, resp *string) error {
+	*resp = s.Impl.GetName()
+	return nil
```
yogeshbdeshpande, line 32: If I am not mistaken, GetName can return a "" string which is an Error, however here without checking for null string, we are returning a nil which means no error?

common/policy_plugin.go
```diff
@@ -0,0 +1,252 @@
...
+
+func (s *PolicyEngineServer) GetName(args interface{}, resp *string) error {
+	*resp = s.Impl.GetName()
+	return nil
+}
+
+func (s *PolicyEngineServer) Init(args PolicyEngineParams, resp *interface{}) error {
+	return s.Impl.Init(args)
+}
+
```
yogeshbdeshpande, line 134: Might be a good idea for Functions beneath Line 134 to have a one liner comment, as to from where they fetch the set, like Policy, Claims or Attestation Results.

endorsement/manager.go
```diff
@@ -0,0 +1,63 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package endorsement
+
+import (
+	"github.com/hashicorp/go-plugin"
+
+	"veraison/common"
+)
+
```
yogeshbdeshpande, line 11: A one liner comment about manager might be helpful here.

plugins/sqliteendorsements/main.go
yogeshbdeshpande, line 99: Might be worth adding a polish comment, as this methid improvement is a Work In Progress"

yogeshbdeshpande (approved):

common/evidenceextractor_plugin.go
setrofim: Here we're just forwarding the result form the implementation `GetName`. Since calling the underlying implementation doesn't result in an error, this shouldn't either. It is part of the definition of `GetName` that it does not error. If `""` is returned it will simply be forwarded to the calling code, as if the calling code invoked `Impl.GetName()` directly. It would then be up to the calling code to handle this case.

common/evidenceextractor_plugin.go
setrofim: Yes, it separates the declaration from the following code block (e.g. see also line 98 above).

common/policy_plugin.go
setrofim: This is documented as part of `PolicyEngine` interface, which this is just forwarding. This is also part of the internal plugin implementation and is not intended to be use by anyone else.

common/evidenceextractor_plugin.go
yogeshbdeshpande: Thanks for the clarification, please resolve the comment then !

common/policy_plugin.go
yogeshbdeshpande: Ok, Thanks then please do resolve the comment!

endorsement/manager.go
setrofim: Ah, missed that. Now added. Thanks!

plugins/sqliteendorsements/main.go
setrofim: I've remove d the comment.

PR #15, opened 2021-02-19
add doc to capture assumptions about token usage

SimonFrost-Arm:


docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
+_Attestation Token use cases & assumptions_
```
thomas-fossati, line 1:
```suggestion
## Attestation Token use cases & assumptions
```

thomas-fossati, line 1: general comment: should we align to RATS glossary and call the attestation token "Evidence" instead?

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+The following captures what is assumed about the nature of the attestation tokens received for verification.
+
+A token is presented as a single 'blob' (byte stream).
+
+Tokens are presented individually. If a batch mode submission is required, that is the responsibility of an API tier.
```
thomas-fossati, line 6:
```suggestion
Tokens are presented individually. If a batch mode submission is required, it is the responsibility of an API tier to pass them for verification one by one.
```

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+
+Encoding in transit is the responsibility of the API tier, the verification pipeline works on a byte stream.
```
thomas-fossati, line 8: It'd be nice if we say what kind of encoding are we talking about here?  My guess is any kind of lossless compression (e.g., what is described by HTTP's `Content-Encoding` header) but you may have another thing in mind?

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+
+The verification pipeline will receive some context information which will indicate or help to deduce the nature of the token and hence which plugins will be needed in the verification pipeline. In a dedicated deployment this may be hardwired, in an environment supporting multiple token types, this is likely to come from the API tier.
+
+A token may have an encrypted body. Assuming the correct decryption key is available, the body will be decrypted and then passed into the pipeline in the same manner as an unencrypted token.
```
thomas-fossati, line 12: I think the fact that the token was encrypted should be passed on to the verification step because this may be relevant to the verification policy.  (E.g., for privacy considerations)

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+
+A presented token blob may actually contain multiple tokens, each of which will be handled e.g. deserialised and have signing checked, independently such that an aggregation of all evidence contained is extracted and presented for appraisal
```
thomas-fossati, line 14:
```suggestion
A presented token blob may actually contain multiple tokens, each of which will be handled - e.g., deserialised and have signing checked - independently such that an aggregation of all evidence contained is extracted and presented for appraisal
```

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+
+Multiple token models include both concatenation (e.g a chain of certs packaged serially) or embedded (e.g. the value of a claim within one token may be a serialised independent token)
```
thomas-fossati, line 16:
```suggestion
Bundled token models include both concatenation (e.g a chain of certs packaged serially) and nesting (e.g. the value of a claim within one token may be a serialised independent token)
```

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+
+The set of endorsements / reference values required to appraise the evidence in a token is assumed to have been provisioned to the verifier service prior to the token being presented. i.e. the verifier cannot distinguish between a bad value in a claim and a 'missing' value.
```
thomas-fossati, line 18: There may be use cases where the endorsement is independently signed and inlined in the token.

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+Encoding in transit is the responsibility of the API tier, the verification pipeline works on a byte stream.
+
+The verification pipeline will receive some context information which will indicate or help to deduce the nature of the token and hence which plugins will be needed in the verification pipeline. In a dedicated deployment this may be hardwired, in an environment supporting multiple token types, this is likely to come from the API tier.
+
+A token may have an encrypted body. Assuming the correct decryption key is available, the body will be decrypted and then passed into the pipeline in the same manner as an unencrypted token.
+
+A presented token blob may actually contain multiple tokens, each of which will be handled e.g. deserialised and have signing checked, independently such that an aggregation of all evidence contained is extracted and presented for appraisal
+
+Multiple token models include both concatenation (e.g a chain of certs packaged serially) or embedded (e.g. the value of a claim within one token may be a serialised independent token)
+
```
thomas-fossati, line 17: I don't understand this bit of the sentence: "i.e. the verifier cannot distinguish between a bad value in a claim and a 'missing' value."

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,21 @@
...
+The set of endorsements / reference values required to appraise the evidence in a token is assumed to have been provisioned to the verifier service prior to the token being presented. i.e. the verifier cannot distinguish between a bad value in a claim and a 'missing' value.
+
+The exception to the above might be where the R.V. is held in an external service which the verifier must consult. In this case, the prior rule applies to that external service.
```
thomas-fossati, line 20:
```suggestion
The exception to the above might be where the endorsement / reference values are held in an external service which the verifier must consult. In this case, the prior rule applies to that external service.
```

docs/Token_Assumptions.md
SimonFrost-Arm: I think I'd prefer to leave it this way as I think it's more approachable for someone new to the area.

docs/Token_Assumptions.md
SimonFrost-Arm: I deliberately generalised because the API call could be in all kinds of context e.g. local API call rather than via REST

docs/Token_Assumptions.md
SimonFrost-Arm: Agree that noting it was encrypted is a useful piece of state, but I wanted to capture that encrypted tokens are supported and that decrypting them is seen as a separate operation.

docs/Token_Assumptions.md
SimonFrost-Arm: Good point, I'll refresh

docs/Token_Assumptions.md
SimonFrost-Arm: I was trying to underline the nature of the workflow: we assume that there isn't a 'live' connection between the use of the attested environment and the supply chain values that match that environment. Thus if the verifier does not find a RV match, there could be two explanations: bad value or missing RV, it's not within the verifier scope to distinguish the two.

docs/Token_Assumptions.md
SimonFrost-Arm: Would there be a scenario where endorsements (using strict definition) would be subject to a live lookup? Seems a bit delicate given the extra checking required.

docs/Token_Assumptions.md
SimonFrost-Arm: Done

docs/Token_Assumptions.md
thomas-fossati: I think this depends on the API exposed by the Endorser.  But we don't make any assumption in the Verifier: in fact, the Endorser is masked by a couple of interfaces (one to the endorsement "store", the other to the trust anchor "store") and the related plugins could do a lot of crazy stuff to collect the ref-values that we don't really care about :)

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,20 @@
...
+
+Encoding in transit is the responsibility of the API tier, the verification pipeline works on a byte stream.
+
+The verification pipeline will receive some context information which will indicate or help to deduce the nature of the token and hence which plugins will be needed in the verification pipeline. In a dedicated deployment this may be hardwired, in an environment supporting multiple token types, this is likely to come from the API tier.
+
+A token may have an encrypted body. Assuming the correct decryption key is available, the body will be decrypted and then passed into the pipeline in the same manner as an unencrypted token.
+
+A presented token blob may actually contain multiple tokens, each of which will be handled - e.g., deserialised and have signing checked - independently such that an aggregation of all evidence contained is extracted and presented for appraisal
+
+Bundled token models include both concatenation (e.g a chain of certs packaged serially) and nesting (e.g. the value of a claim within one token may be a serialised independent token)
```
thomas-fossati, line 16:
```suggestion
Bundled token models include both concatenation (e.g., a chain of certs packaged serially) and nesting (e.g., the value of a claim within one token may be a serialised independent token)
```

docs/Token_Assumptions.md
```diff
@@ -0,0 +1,20 @@
...
+
+The set of endorsements / reference values required to appraise the evidence in a token is assumed to have been provisioned to the verifier service prior to the token being presented. i.e. the verifier cannot distinguish between a bad value in a claim and a 'missing' value. The only exception to this is if an endorsement is included inline within the token and independently verifiable.
```
thomas-fossati, line 18:
```suggestion
The set of endorsements / reference values required to appraise the evidence in a token is assumed to have been provisioned to the verifier service prior to the token being presented. I.e., the verifier cannot distinguish between a bad value in a claim and a 'missing' value. The only exception to this is if an endorsement is included inline within the token and independently verifiable.
```

docs/Token_Assumptions.md
thomas-fossati: OK, thanks.  Now I get the intent.  I'm worried that making this explicit in this context may baffle the simple minds (like myself!) without adding real substance to what we are trying to say?

SimonFrost-Arm: Don't want to confuse so will remove this sentence

thomas-fossati (approved):
thanks for addressing my comments!

docs/Token_Assumptions.md
SimonFrost-Arm: removed to calm cerebral squalls

PR #41, opened 2021-03-31
Fixes #39 GetParam API to use strong typing

### LINKED ISSUE #39: For GetQueryParam API consider using strong typing
I'm not sure of the reason why `input` is of type `interface{}`.  Better if we can use a stricter typing.

_Originally posted by @thomas-fossati in https://github.com/veraison/veraison/pull/37#discussion_r603392092_

yogeshbdeshpande:
Use strong type for GetQueryParam API instead of interface{}

### DIFF
--- a/plugins/arangodbendorsements/arangodb.go
+++ b/plugins/arangodbendorsements/arangodb.go
@@ -142,9 +142,9 @@ func (e *ArangoStore) IsInitialised() bool {
 	return e.isInitialised
 }
 
-// GetQueryParam returns a suitable query parameter based on supplied input
-func (e *ArangoStore) GetQueryParam(input string) (string, error) {
-	switch input {
+// GetQueryParam returns a suitable query parameter based on supplied qParamID
+func (e *ArangoStore) GetQueryParam(qParamID string) (string, error) {
+	switch qParamID {
 	case HW:
 		return e.dbparams.HwCollection, nil
 	case SW:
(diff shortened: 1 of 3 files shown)


plugins/arangodbendorsements/arangodb.go
```diff
@@ -119,8 +119,8 @@ func (e *ArangoStore) IsInitialised() bool {
 }
 
 // GetQueryParam returns a suitable query parameter based on supplied input
-func (e *ArangoStore) GetQueryParam(input interface{}) (string, error) {
-	switch input.(CollectionType) {
+func (e *ArangoStore) GetQueryParam(input string) (string, error) {
```
thomas-fossati, line 122: maybe instead of `input`, which has no particular meaning, use `queryParamID` or similar?

yogeshbdeshpande: ID looks a little odd as it is a string, may be simple to name it as qParam ?

thomas-fossati: the question is: what is this `input` parameter?  is it a query parameter (qParam)? or is it a unique identifier for the query parameter (qParamID)?

yogeshbdeshpande: Resolved it by renaming input to qParamID

thomas-fossati (approved):

PR #44, opened 2021-04-22
Implement global configuration and endorsement tooling.

setrofim:


common/query.go
```diff
@@ -6,13 +6,102 @@ package common
...
+	default:
+		return "", fmt.Errorf("unexpected type for '%s'; must be a string; found: %T", name, v)
+	}
+
+	return result, nil
+}
+
+// GetStringSlice returns the value for the specified name as a string slice,
+// or an error if the value could not be interpreted as such.
+func (a QueryArgs) GetStringSlice(name string) ([]string, error) {
```
thomas-fossati, line 54: I'd have called this `GetAsStringArray` (I find the use of `Slice` in this context a bit confusing, but it may be me...)

common/query.go
```diff
@@ -6,13 +6,102 @@ package common
...
+			return "", fmt.Errorf("unexpected type for '%s'; must be a string", name)
+		}
+	default:
+		return "", fmt.Errorf("unexpected type for '%s'; must be a string; found: %T", name, v)
+	}
+
+	return result, nil
+}
+
+// GetStringSlice returns the value for the specified name as a string slice,
```
thomas-fossati, line 52: maybe we should say that a `nil` `[]string` is an ok return value.

common/query.go
```diff
@@ -21,6 +110,11 @@ type QueryResult []interface{}
 // QueryResult contain the query's matches.
 type Query func(QueryArgs) (QueryResult, error)
 
+// QueryAdder defines the adder function signature. These functins are used to
```
thomas-fossati, line 113: s/functins/functions/

common/query.go
```diff
@@ -21,6 +110,11 @@ type QueryResult []interface{}
...
+// add ensorsements to the store from the proveded QueryArgs, or alternatively,
```
thomas-fossati, line 114: s/proveded/provided/

plugins/sqliteendorsements/main.go
```diff
@@ -163,7 +154,136 @@ func (e *SqliteEndorsementStore) GetSoftwareComponents(args common.QueryArgs) (c
...
+	err := row.Scan(&typ, &signerID, &version)
+	if err == nil {
+		if typ != component.Type || signerID != component.SignerID || version != component.Version {
+			return fmt.Errorf(
+				"component with measurement \"%s\" is already registered",
+				component.Measurement,
+			)
+		}
+
+		// This exact component extry already exists; short-circuit here, indicating success.
```
thomas-fossati, line 209: s/extry/entry/ (fantastic typo btw!)

thomas-fossati (approved):
LGTM! I have left just a few non-blocking comments.

common/config.go
```diff
@@ -0,0 +1,123 @@
...
+	c.EndorsementStoreName = c.viper.GetString("endorsements.store_name")
+	c.EndorsementStoreParams = c.viper.GetStringMap("endorsements.store_params")
+
+	return nil
+}
+
+func (c Config) validate() error {
+	for _, key := range expectedConfigKey {
+		if !c.viper.IsSet(key) {
+			return fmt.Errorf("key \"%s\" not set in configuration", key)
```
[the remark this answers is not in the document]
setrofim: Oh, that's awesome! Thank you.

common/query.go
setrofim: I'm being consistent with Viper here (https://github.com/spf13/viper/tree/v1.7.1#getting-values-from-viper), but can change if you'd prefer.

common/config.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// A set of default paths that can be specified for Config.Init().
+var DefaultConfigPaths = []string{
+	"$HOME/veraison",
+	".",
+}
+
+var expectedConfigKey = []string{
+	"plugin_locations",
```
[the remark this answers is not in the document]
setrofim: Hm, good point. Will make the change.

common/query.go
thomas-fossati: cool, thanks for the pointer.

PR #49, opened 2021-06-23
Policy tooling

setrofim:
Add tooling to manipulate policy store.


cmd/policy/README.md
```diff
@@ -0,0 +1,50 @@
...
+Add policies from my_policies.zip for tenant 1, overwriting any existing policies.
+
+   ./policy -c /opt/veraison/config/ get -t 3 -o my_policies.zip psa dice
+
+Get policies for psa and dice token formats for tenant 3, using Veraison
+configuration from /opt/veraison/config/, and write them to my_policies.zip.
+
+ ./policy verify -e test/endorsements.json test/evidence-context.json
+
+Run stored policy against input inside test/evidence-context.json, using
```
thomas-fossati, line 23: Which policies are used in this case?  Is there a default tenant or all applies?

cmd/policy/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Policies can be read out from the store into a zip archive. A similar archive
+can be used to add policies to the store. Each policy is contained within a
+top-level directory within the archive, named after the token format. The
+directory contains two files: query_map.json is a JSON serialization of the
+query map -- a mapping indicating how endorsements query parameters can be
+obtained from the token claims; the second file is "rules" and contains the
+rules to be used for evaluating token evidence against endorsements. The format
+of the rules must match what is expected by the Policy Engine configured in a
+particular Veraison deployment (e.g. in REGO format for the OPA engine).
```
thomas-fossati, line 36: where do we encode the tenant in the hierarchy?

cmd/policy/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+	return nil
+}
+
+func runGetCommand(config *common.Config, args []string, pm *policy.Manager, logger *zap.Logger) error {
+	var tenantID int
+	var outPath string
+
+	addFlags := flag.NewFlagSet("add", flag.ExitOnError)
+	addFlags.IntVar(&tenantID, "t", 1, "TenantID to which the policies should be added.")
```
thomas-fossati, line 83: is there a reason why the default tenant for `list` is 0, while the default tenant for `add` is 1?  Or maybe `0` has a special meaning (if so, we should document it)?

cmd/policy/main.go
```diff
@@ -0,0 +1,397 @@
...
+					return fmt.Errorf(
+						"could not remove existing policy for tenantID %d and format %q: %v",
+						tenantID,
+						policy.TokenFormat.String(),
+						err,
+					)
+				}
+			} else {
+				return fmt.Errorf(
+					"policy for tenant 1 and fromat %q already exists; use -f to overwrite",
```
thomas-fossati, line 170: typo: `"policy for tenant %d [...]", tenantID)`

cmd/policy/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+	var policies []*common.Policy
+	for _, format := range formats {
+
+		policy, err := pm.GetPolicy(tenantID, format)
+		if err != nil {
+			return fmt.Errorf(
+				"could not get policy for tenant %d, format %q: %v",
+				tenantID,
+				format.String(),
```
thomas-fossati, line 113: here and in other similar instances: if the type is a stringer -- i.e., it provides a `String()` method  -- you can avoid calling it explicitly (`Errorf` will do it for you.)

cmd/policy/main.go
```diff
@@ -0,0 +1,397 @@
...
+
+verify
+    Run a policy  against the specified endorsements and evidence context. This
+    can be used to test a policy to make sure it behaves as expected against a
+    known input. The following additional arguments are supported:
+
+	[-e ENDORSEMENTS] EVIDENCE_CONTEXT
+
+    where ENDORSEMENTS is the path to JSON-encoded endorsement values, and
+    EVIDENCE_CONTEXT is the path a file containing JSON-serialized
```
thomas-fossati, line 355: typo: "path to a file"?

cmd/policy/main.go
```diff
@@ -0,0 +1,397 @@
...
+    EvidenceContext.
+
+Top-level Options:
```
thomas-fossati, line 358: looks like this should be the place where `-c` and `-d` are documented?

thomas-fossati (commented):
LveryGTM, thanks!

I have left a bunch of small editorial comments.

cmd/Makefile
```diff
@@ -0,0 +1,7 @@
+# Copyright 2021 Contributors to the Veraison project.
+# SPDX-License-Identifier: Apache-2.0
+
+SUBDIR := endosements
```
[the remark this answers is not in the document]
setrofim: Argh, yeah, it's actually a typo in the dir name. Will fix.

cmd/policy/README.md
```diff
@@ -0,0 +1,50 @@
...
+List all policies in Veraison store. This will output a list of tenant ID and
+format name pairs.
+
+    ./policy -t 2 list
+
+Same as above but only list policies for tenant ID 2.
+
+    ./policy add -f my_policies.zip
+
+Add policies from my_policies.zip for tenant 1, overwriting any existing policies.
```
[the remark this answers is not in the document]
setrofim: > should the tenant be explicitly passed on the command line?

I'm treating single tenancy as the common case (at least, as far as the tooling is concerned -- it's there primarily to help with development / testing of initial deployment), hence there is a default assumption of tenant 1 if the flag is not specified -- that way you don't have to worry about it for single-tenant case.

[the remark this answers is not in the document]
setrofim: > if this overrides the existing policies, isn't set a more fitting name for the command rather than add?

Hm, makes sense. Will rename.

cmd/policy/README.md
setrofim: Defaults to tenant 1. I will update the docs to make that clearer. Thanks!

cmd/policy/README.md
setrofim: We don't. Tenant ID is an internal artefact of a particular deployment, not the Policy (e. g. a Policy Provider may not be aware of what their tenant ID is). Instead, Tenant ID is specified as part of the evidence context.

cmd/policy/main.go
setrofim: Yes -- for `list`, if the tenant is not specified, then the default assumption is "all tenants" (i.e. list everything); for `add`, the assumption is "the default tenant" (it doesn't make sense to add a policy to all tenants). I've updated the README with a note to clarify this.

cmd/policy/main.go
setrofim: D'oh. Thanks!

cmd/policy/main.go
setrofim: Oh, cool! Thanks!

cmd/policy/main.go
setrofim: Yup, thanks!

cmd/policy/main.go
setrofim: These are generated by `flag.PrintDefaults()` in the `main()`.

cmd/policy/README.md
thomas-fossati: thanks for the explanation!

cmd/policy/main.go
thomas-fossati: ah!  my bad, sorry for the silly comment.

cmd/policy/README.md
```diff
@@ -0,0 +1,55 @@
...
+    ./policy list
+
+List all policies in Veraison store. This will output a list of tenant ID and
+format name pairs.
+
+    ./policy -t 2 list
+
+Same as above but only list policies for tenant ID 2.
+
+    ./policy add -f my_policies.zip
```
thomas-fossati, line 17: If I am not mistaken this should say `set` now?

setrofim: Do'h, missed that. Now fixed.

cmd/policy/main.go
```diff
@@ -0,0 +1,398 @@
...
+					return fmt.Errorf(
+						"could not remove existing policy for tenantID %d and format %q: %v",
+						tenantID,
+						policy.TokenFormat,
+						err,
+					)
+				}
+			} else {
+				return fmt.Errorf(
+					"policy for tenant %d and fromat %q already exists; use -f to overwrite",
```
thomas-fossati, line 170: typo: fromat

thomas-fossati (approved):
super!

PR #50, opened 2021-07-02
add sketch of attestation results format

thomas-fossati:
* CDDL sketch that tries to capture our conversation with JSON and CBOR
  serialisations
* framework for checking integrity of CDDL for both JSON and CBOR
* framework for adding samples and checking them against the current
  CDDL


docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,43 @@
...
+attestation-results = {
+  ar.status => $ar-status
+  ar.trust-vector => ar-trust-vector
+  ? ar.raw-evidence => ar-bytes
+  ar.timestamp => ar-timestamp
+  ? ar.endorsed-claims => ar-endorsed-claims
+}
+
+ar-trust-vector = {
+  ar.hw-authenticity => $ar-status
```
SimonFrost-Arm, line 16: I can see this status working for most scenarios. Is there a case where values may be needed instead e.g. a certification level, an overall SW security epoc? Appreciate that those could be in the extension bucket but wanted to raise the query in case these results should be named ...-status for clarity there

SimonFrost-Arm: Scratch that, just read the end of the file has the -details sections!

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,43 @@
...
+}
+
+ar-sw-details = {
+  * $$ar-sw-details-extension
+}
+
+ar-certification-details = {
+  * $$ar-certification-details-extension
+}
+
```
SimonFrost-Arm, line 42: Does up-to-dateness need a details section? e,g, a timestamp or version number for up to date release indentity?

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,43 @@
+start = attestation-results
+
+$ar-status /= ar.status-failure
+$ar-status /= ar.status-success
+$ar-status /= ar.status-unknown
+
+attestation-results = {
+  ar.status => $ar-status
+  ar.trust-vector => ar-trust-vector
+  ? ar.raw-evidence => ar-bytes
```
SimonFrost-Arm, line 10: Is the intent here just to be bytes or a structured replay of evidence (potentially with 'checked' flags)? RP may well have additional appraisal path for fields which require RV or Endorsement checking beyond scope of verifier. e.g. software workload measurements on a verified platform

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,43 @@
+start = attestation-results
+
+$ar-status /= ar.status-failure
+$ar-status /= ar.status-success
+$ar-status /= ar.status-unknown
+
+attestation-results = {
+  ar.status => $ar-status
+  ar.trust-vector => ar-trust-vector
```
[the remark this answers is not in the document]
thomas-fossati: Good point!  I missed the `?` :-(

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,43 @@
...
+  ar.config-integrity => $ar-status
+  ar.runtime-integrity => $ar-status
+  ar.certification-status => $ar-status
+  * $$ar-trust-vector-extension
+}
+
+ar-endorsed-claims = {
+  ar.hw-details => ar-hw-details
+  ar.sw-details => ar-sw-details
+  ar.certification-details => ar-certification-details
```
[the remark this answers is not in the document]
thomas-fossati: As far as my interpretation goes (caveat! ;-)) the certification status in the trust vector is the derived part, while this here is just the copy-paste from the certification record that is obtained from the "endorsement" path.

(On a more syntactical / layman level of musing, I think all these guys should be optional and the top-level wrapped in a `non-empty<>` macro.)

docs/dm/attestation-results/attestation-results.cddl
thomas-fossati: The idea was to use the sw-details inside the endorsed claims block to stash that kind of info.

docs/dm/attestation-results/attestation-results.cddl
thomas-fossati: We debated this quite extensively, and the rationale for keeping it a verbatim copy rather than a re-structured thing is that this way we avoid adding extra interface surface on a per-token-format basis, and therefore avoid any coupling that is not strictly required.  Note that, since the format is extensible, this doesn't preclude a closer RP-Verifier relationship where the Verifier takes care of making Evidence into something easier to consume by the RP.

SimonFrost-Arm: I can see the argument both ways, but was inspired by the spirit of veraison taking care of all token specific details and hence being in an ideal position to produce a highly readable / processable form of the evidence for an RP (it's available as an internal structure after all)! Maybe there should be an optional ar.found-evidence or similar placeholder to support further discussion.

thomas-fossati: If we want to add it I'd make it a "Veraison feature" using a (TBD) extension point rather than a core feature that all verifiers implementing this interface need to provide, i.e.:
```
; to be added to the attestation-results map
$$attestation-results-extension //= (
  ar-veraison.found-evidence => ar-veraison-found-evidence
)

ar-veraison-found-evidence = {
  $ar-label => any
}
```

SimonFrost-Arm: Understood - I had missed the nuance that the CDDL work was for a standard implementation rather than just veraison

SimonFrost-Arm (approved):

docs/dm/attestation-results/attestation-results.cddl
thomas-fossati: cool, I'm adding the (currently) missing extension point then!

setrofim (approved):
LGTM! :+1:

docs/dm/attestation-results/attestation-results.cddl
thomas-fossati: Done in 9717e6a

thomas-fossati:
rebased on tip to pull in Sergei's 3af2d65

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,44 @@
...
+}>
+
+ar-hw-details = {
+  + ar-label => any
+}
+
+ar-sw-details = {
+  + ar-label => any
+}
+
```
yogeshbdeshpande, line 39: Should we also have a config-details section here! I recollect in many attestation use cases, configuration becomes a key element of Attestation ?

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,44 @@
...
+  ar.config-integrity => $ar-status
+  ar.runtime-integrity => $ar-status
+  ar.certification-status => $ar-status
+  * $$ar-trust-vector-extension
+}
+
+ar-endorsed-claims = non-empty<{
+  ? ar.hw-details => ar-hw-details
+  ? ar.sw-details => ar-sw-details
+  ? ar.certification-details => ar-certification-details
```
yogeshbdeshpande, line 29: Should there not be an extension point to the ar-endorsed-claims in-line with trust vector extension ?

docs/dm/attestation-results/attestation-results.cddl
thomas-fossati: I am not sure that configuration endorsements would fit in to the SW category.  It looks to me that if we want to have them, we should make a placeholder in the ar-endorsed-claims level?

docs/dm/attestation-results/attestation-results.cddl
thomas-fossati: makes sense!

docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -0,0 +1,44 @@
...
+
+$ar-status /= ar.status-failure
+$ar-status /= ar.status-success
+$ar-status /= ar.status-unknown
+
+attestation-results = {
+  ar.status => $ar-status
+  ? ar.trust-vector => ar-trust-vector
+  ? ar.raw-evidence => ar-bytes
+  ar.timestamp => ar-timestamp
```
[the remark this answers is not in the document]
thomas-fossati: I think it is OK as it is for now.  If this should become a serious source of shock we can reshuffle.

docs/dm/attestation-results/attestation-results.cddl
yogeshbdeshpande: Yes I meant at endorsed level only.

I mean at ar-endorsed-claims = non-empty<{
? ar.config-details => ar-config-details
}

and then expand the config. like you did for other details ?

thomas-fossati: > Yes I meant at endorsed level only.

Cool, makes sense. I'll add it.

yogeshbdeshpande: Thank You!

docs/dm/attestation-results/attestation-results.cddl
yogeshbdeshpande: Thank You!

docs/dm/attestation-results/attestation-results.cddl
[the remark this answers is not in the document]
yogeshbdeshpande: Its just that attestation result and its time stamp where the results are applicable are more aligned together. No hard views, as I said earlier :)

yogeshbdeshpande (approved):

docs/dm/attestation-results/attestation-results.cddl
SimonFrost-Arm: It's possible that config items will fit into the same bucket as software items e.g. the config item is likely to be a signed blob which can be measured as loaded. Only a description / name would distinguish from a similarly distributed excutable component

yogeshbdeshpande: As I understand, TCG and other specifications clearly see a difference in SW and Config. Better to keep them separate, though I agree the actual signed blob may look very similar semantically and the steps to verify may also be similar

PR #52, opened 2021-07-07
Clarifying README.md

yogeshbdeshpande:
Attempt to clarify and tidy a bit top level page


README.md
```diff
@@ -18,10 +18,10 @@ Project _VERAISON_: **VER**ific**A**t**I**on of atte**S**tati**ON**
...
+Given the above, it can be challenging to build just one Verification Service solution which can address all types of deployments for a technology that needs to produce Attestation reports to prove its trustworthiness. It then implies that each deployment needs a custom service, however there is a significant software barrier and hence cost of entry to establishing such a system to be used in a secure manner. Veraison aims to provide consistency and convenience to solving this problem by building the software components that can be used to build Attestation Verification Services. The Veraison project will also build some reference implementations in order to prove the integration principles of those components and to provide a convenient starting point for some implementations.
 
 For further details, see the documentation [TBD]
```
yogeshbdeshpande, line 26: is it a good time to add a link the existing documents we have ?

README.md
```diff
@@ -18,10 +18,10 @@ Project _VERAISON_: **VER**ific**A**t**I**on of atte**S**tati**ON**
...
+Given the above, it can be challenging to build just one Verification Service solution which can address all types of deployments for a technology that needs to produce Attestation reports to prove its trustworthiness. It then implies that each deployment needs a custom service, however there is a significant software barrier and hence cost of entry to establishing such a system to be used in a secure manner. Veraison aims to provide consistency and convenience to solving this problem by building the software components that can be used to build Attestation Verification Services. The Veraison project will also build some reference implementations in order to prove the integration principles of those components and to provide a convenient starting point for some implementations.
```
SimonFrost-Arm, line 24: A suggestion instead of your 'however' change:   s/however there is/which creates/

It then implies that each deployment needs a custom service, which creates a a significant software barrier...

yogeshbdeshpande: ok, this is also fine, I will change it!

README.md
```diff
@@ -18,10 +18,10 @@ Project _VERAISON_: **VER**ific**A**t**I**on of atte**S**tati**ON**
...
+Attestation is the system by which an entity produces evidence about itself that another party can use to evaluate the trustworthiness of that entity. A critical part of establishing trust based on that evidence is Verifying what is presented as evidence against a set of known elements to ensure it meets an established policy to prove trustworthiness. The act of Verification will often require comparing evidence against reference values or checking cryptographic signing by a known trust anchors. This is a complex task to be performed by an arbitrary user of a system and it is often delegated to a trusted Verification Service instead. The nature of the Service will depend upon specific needs - it may be a local software component or it may reside as a network or internet accessible component. The actual nature of the Service will depend upon the actual deployment. The Verification Service will need up to date data for the process of appraising the evidence presented to it. This must be obtained from authoritative sources, like the supply chain, which normally implies establishing business relationships between the Verification Service and those sources.
```
thomas-fossati, line 21: I don't think "known elements" is the right word combo here.

Maybe:

_A critical part of establishing trust based on that evidence is verifying that the claims it conveys match "known-good values" for each relevant trust metric indicator according to the established appraisal policy._

thomas-fossati: Maybe "typically, a supply chain actor" rather than "like the supply chain"

yogeshbdeshpande: > I don't think "known elements" is the right word combo here.
> 
> Maybe:
> 
> _A critical part of establishing trust based on that evidence is verifying that the claims it conveys match "known-good values" for each relevant trust metric indicator according to the established appraisal policy._

This is what has been already covered in the subsequent sentence, hence we do not want to repeat it here. This is a high level sentence for setting context and further clarification of comparison follows in the next sentence

thomas-fossati: Ah, right!  OK, so I think I prefer the previous version then.  The thing is I have a problem defining what a "known element" is.

yogeshbdeshpande: > Maybe "typically, a supply chain actor" rather than "like the supply chain"

do we want to restrict authoritative source only to supply chain, if yes, then may be typically is a right word.

If we want to have a broader scope for authoritative source, example or like the is a correct word to use ?

WDYT?

yogeshbdeshpande: > Ah, right! OK, so I think I prefer the previous version then. The thing is I have a problem defining what a "known element" is.



> Ah, right! OK, so I think I prefer the previous version then. The thing is I have a problem defining what a "known element" is.

The problem with the original version is, it directly jumps from Evidence to policy without a "logical flow"
Hence this is a Bridge! Happy to swap a further improved word for "Known Elements" :)

thomas-fossati: > > Maybe "typically, a supply chain actor" rather than "like the supply chain"
> 
> do we want to restrict authoritative source only to supply chain, if yes, then may be typically is a right word.

I don't think "typically" is too stringent.  It just indicates what the expected norm is while still allowing deviations.

yogeshbdeshpande: Fine,  I will change this !

SimonFrost-Arm: Adding the bridge does make the sentence rather long. If the use of policy is distracting because that's also the name of an artifact, perhaps we should swap 'a policy' for 'rules' which is a more abstract expression. As you say, the following sentences go into more detail

yogeshbdeshpande: I think, I found the right sentence :)

A critical part of establishing trust based on evidence is Verifying 
the evidence as per the established policy of verification to prove trustworthiness.

I think, this is clear. Comments from both of you Welcome on the above sentence!

SimonFrost-Arm: Nearly, but I think the double use of evidence can be avoided thus:
A critical part of establishing trust is Verifying that evidence, as per the established policy of verification, to prove trustworthiness.

thomas-fossati: I am not sure this would be an improvement over what we already have.

Maybe:

_A critical part of establishing trust based on that evidence is verifying what is presented against a set of appraisal rules to determine whether the reported trust metrics match the desired state._

SimonFrost-Arm: I like that, but how about just:    "A critical part of establishing trust based on that evidence is verifying what is presented against a set of appraisal rules"

yogeshbdeshpande: @thomas-fossati : I will go with improving what Simon already had on the page and I think, the one he mentioned above makes sense. 

I would prefer the Introductory README text to be simple and would prefer to have your statement in one of our internal documentation!

README.md
yogeshbdeshpande: Reviewed and we need to align internal documents, before we can add the reference here!

README.md
thomas-fossati: > I like that, but how about just: "A critical part of establishing trust based on that evidence is verifying what is presented against a set of appraisal rules"

That is correct statement, but it doesn't say what the end goal is, which I think is the crucial thing to understand at this level of presentation.

yogeshbdeshpande: A critical part of establishing trust is Verifying that evidence, as per the established policy of verification, to prove trustworthiness.

This has the simplicity and completeness what we initially had envisaged?

thomas-fossati: > This has the simplicity and completeness what we initially had envisaged?

I am not sure :-) I think to be complete it needs to declare what the end goal is, i.e. something along the lines of _"determine whether the reported trust metrics match the desired device state"_

yogeshbdeshpande: The completeness happens when you read the statement in conjunction with the next statement.
A critical part of establishing trust is Verifying that evidence as per the established policy of verification, to prove trustworthiness.  The act of Verification will often require comparing evidence against reference values or checking cryptographic signing by a known trust anchors.

Here the end goal is to prove :trustworthiness.
You take established policy as an input to verification
Actual verification step is by comparing to reference values.

In my view, it is simple, clear and complete!

thomas-fossati: If I had to read and make sense of it without my last 2 years of brain massage via IETF and TCG, I'd ask what is "trustworthiness" and would not find an answer.

yogeshbdeshpande: Hmmm: one can use this in English term also:
Trustworthy: able to be relied on as honest or truthful.

thomas-fossati: I appreciate that the anthropomorphisation of cold-blooded Turing machines is something we tend to do, but (when possible) we should avoid the temptation :)

yogeshbdeshpande: Let's discuss this further in our Veraison Open Source Meeting. If we still have no consensus, then  I prefer we abandon this PR, instead of going in circles!

thomas-fossati: I think I have made a concrete proposal that would allow the conversation to converge :-) The gist is we provide a concrete characterisation of the (for me) extremely abstract _trustworthiness_ term as _"the measure of the deviation between the reported and the desired integrity state of the attesting device."_

yogeshbdeshpande: Thanks and I already gathered that:
However my points are as below:

1) The term trustworthiness is already used in the existing REAMDE document( which we all reviewed then) right from first statement. 

2) I think, it is quite obvious to a general reader also, what this term implies. In my view it is not a super complex terminology, which cannot be inferred in the flow of sentence!

3) The extra explanation that you provide for this makes the sentence look very long and winding, which may be OK for deep technical documentation inside pages, but prefer not to have it on the introductory page.

4) As  a middle ground, what we can consider is a reference or a glossary in the bottom of the top README to refer to its deep definition (what you have) OR a clickable item for trustworthiness which will take the reader to a definition page where it can view the detail definition if required ?

WDYT?

thomas-fossati: strongly disagree on 2.
mildly disagree on 3 -- but I can see the point of heavy prose discouraging the reader.
strongly agree on 4. (a footnote would work)

SimonFrost-Arm: Apologies for dropping out of the conversation for a while, I've been offline due to IT problems.
I don't think we'll resolve this satisfactorily here, let's postpone to a live conversation.

yogeshbdeshpande: @SimonFrost-Arm Works for me! Lets discuss during our Veraison Open Source Meeting!

yogeshbdeshpande:
Will not address at the moment!

Issue #27, opened 2021-03-08
feature: attestation result format


thomas-fossati:
### Is your feature request related to a problem? Please describe

At present, the [challenge-response API](https://github.com/veraison/veraison/blob/main/docs/api/README.md) returns the attestation result in form of "naked" JSON object with a custom / free-form format.

There are two aspects to address here:
1. moving from custom / free-form to something that is more stable;
2. allow attestation results to be signed by the attester.

We should move towards the model defined by the RATS architecture [bow tie diagram](https://www.ietf.org/archive/id/draft-ietf-rats-architecture-10.html#figure-9):

```
                Evidence           Attestation Results
.--------------.   CWT                    CWT   .-------------------.
|  Attester-A  |------------.      .----------->|  Relying Party V  |
'--------------'            v      |            `-------------------'
.--------------.   JWT   .------------.   JWT   .-------------------.
|  Attester-B  |-------->|  Verifier  |-------->|  Relying Party W  |
'--------------'         |            |         `-------------------'
.--------------.  X.509  |            |  X.509  .-------------------.
|  Attester-C  |-------->|            |-------->|  Relying Party X  |
'--------------'         |            |         `-------------------'
.--------------.   TPM   |            |   TPM   .-------------------.
|  Attester-D  |-------->|            |-------->|  Relying Party Y  |
'--------------'         '------------'         `-------------------'
.--------------.  other     ^      |     other  .-------------------.
|  Attester-E  |------------'      '----------->|  Relying Party Z  |
'--------------'                                `-------------------'
```

### If possible, describe a solution

1. Define the format of Veraison attestation results
    1. Document it publicly (e.g., in the relevant OpenAPI component's schema, and/or in an Internet draft);
    2. Register the associated MIME format (e.g., `application/vnd.veraison-attestation-result+json`);
    3. Adapt codebase.
3. Define a signed envelope that can be used to carry it (maybe we could use JWT and CWT always, with optional signature)

### Describe alternatives you have considered

No alternatives have been considered for the moment.

## Additional context

* [Consideration for using "none" algorithm in JWT](https://tools.ietf.org/html/rfc8725#section-3.2)
* [Unprotected CWTs](https://tools.ietf.org/html/draft-birkholz-rats-uccs)

SimonFrost-Arm:
[...]

PR #54, opened 2021-07-13
add an appraisal policy identifier

thomas-fossati:
As the title says, this PR adds a mandatory identifier of the appraisal policy of type text.

(This also moves the base for the extension code points allocation to 100 to avoid overlaps -- at least for a while.)

(Working towards #27)


docs/dm/attestation-results/attestation-results.cddl
```diff
@@ -10,6 +10,7 @@ attestation-results = {
   ? ar.raw-evidence => ar-bytes
   ar.timestamp => ar-timestamp
   ? ar.endorsed-claims => ar-endorsed-claims
+  ar.appraisal-policy-id => text
```
yogeshbdeshpande, line 13: Should we leave it just as a text or a type choice with restricted set ?
like a resource identifier or something similar?

thomas-fossati: what do you mean exactly by "resource identifier"?

yogeshbdeshpande: I meant URI, here!

thomas-fossati: The question is whether the receiver needs to make sense of the returned policy-id (in which case typing it makes sense), or it is just an opaque handler whose "resolution" is only known to the Verifier.

SimonFrost-Arm: Our intent was human consumption rather than machine so I would favour leaving it as text.

yogeshbdeshpande: Thanks for the clarification. Ok, understood the intent. Fine with me. I will mark the conversation as resolved!

yogeshbdeshpande (approved):

PR #55, opened 2021-07-13
Add 'quiet' param for LoadPlugins

setrofim:
Plugins use a different logger form the rest of the framework due to the
fact that they run in a different process. Add a 'quiet' flag to
LoadPlugins (and expose it via appropriate InializeStore() methods) to
allow suppressing output from the plugin.

(Note such output can always be considered to be "debug", as any actual
issue will be propagated as errors to the main process).

This is primarily to allow suppressing of intermittent output for
testing.

### DIFF
--- a/cmd/endorsements/main.go
+++ b/cmd/endorsements/main.go
@@ -110,16 +110,19 @@ func runListQueriesCommand(
 func runCommand(config *common.Config, command string, args []string, logger *zap.Logger) error {
 	var err error
 
+	quiet := true
 	if config.Debug {
 		if logger, err = zap.NewDevelopment(); err != nil {
 			return err
 		}
 		defer logger.Sync() //nolint
+
+		quiet = false
 	}
 
 	em := endorsement.NewManager()
 	if err := em.InitializeStore(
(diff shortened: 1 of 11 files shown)


cmd/endorsements/main.go
```diff
@@ -110,16 +110,19 @@ func runListQueriesCommand(
...
 		}
 		defer logger.Sync() //nolint
+
+		quiet = false
 	}
 
 	em := endorsement.NewManager()
 	if err := em.InitializeStore(
-		config.PluginLocations, config.EndorsementStoreName, config.EndorsementStoreParams,
+		config.PluginLocations, config.EndorsementStoreName, config.EndorsementStoreParams, quiet,
```
thomas-fossati, line 125: this may be a `config` parameter too, I think?

thomas-fossati (approved):

cmd/endorsements/main.go
setrofim: I considered that, but this really more of internal decision. This flags doesn't really set the logging level, but suppresses output that would be there otherwise. The only reason to do that is for testing (if this flag is false, the logging is set to WARN level).

PR #67, opened 2021-09-10
Add doc describing core assumptions on device modelling

SimonFrost-Arm:
NB: using PlantUML for now just to get something in place.


docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Device evidence classification
+The evidence within an attestation report can be categorised into the following classes. These classes are not an exhaustive set of all types of evidence that may be found within a report but are those necessary for correlating evidence with endorsement / reference value queries.
+
+### Trust Anchor
+All devices are ASSUMED to contain a Trust Anchor (TA). This is normally provisioned into the device at manufacturing time but may also be provisioned to the device on enrolling into a management structure. The TA takes the form of a key or a seed from which an Attestation signing key can be derived. The attestation report will contain information, referred to here as a Trust Anchor ID (TA-ID), which a Verifier can use to locate an endorsement providing information on the public value of the relevant key to use to validate the signature on a report. The Trust Anchor may have unique-to-device or group scope (i.e. multiple devices contain the same trust anchor). The latter case has been used as one solution for privacy protection use cases.
```
thomas-fossati, line 16: in DICE, the trust anchor (i.e., the CA cert that signs the first public key in the DICE hierarchy) might be purely endorsed -- i.e., it may be not necessarily contained inside the device.

Maybe what we really mean here is "signing key for Evidence." and the TA-ID is really an identifier of the Attester (with either instance or group scope) that is bound to that signing key by some means -- for example, in a certificate signed by a CA, in an Endorsement from the supply chain.

docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Hardware Group
+A device MAY contain a Hardware Group ID that is presented in attestation evidence. This is an identifier, provisioned during manufacturing, which can be used to classify the SoC used within the device and will be the same for any devices built using that SoC. Although termed as a hardware ID, in fact this identifier will classify the immutable Root of Trust, being the un-modifiable and un-measured initial BIOS as well as the specific SoC hardware within the SoC. The identifier is a public value and has no security value, but aids classification for a verifier. If no Hardware Group ID is included within the attestation evidence then an equivalent value can be provisioned into the verifier as a correlation to the Trust Anchor ID.
```
thomas-fossati, line 19: the term BIOS looks a bit too specific.  Do we mean "trusted boot block" instead?

docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,76 @@
...
+
+![Groupings](http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/veraison/veraison/main/docs/diags/tokens-model.puml)
+
+Diagram: Known Tokens
+
+![Diagram: Mapping known attestation tokens to the evidence classifications](http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/veraison/veraison/main/docs/diags/known-models.puml)
+
+
+## Firmware Lifecycles
+Where an attestation report contains only a firmware version, no additional data is required from the supply chain and whether the value provided is acceptible is a matter for the evidence appraisal policy.
```
thomas-fossati, line 42: Even though this is not the case for PSA, the SW version could also be a reference value that is checked against the presented Evidence.

yogeshbdeshpande:
Very Nice! I will review it on Tuesday (14th Sept).

thomas-fossati (commented):
Looks good me, thanks!  I have left a few comments inline.

PS: I know in our chests beats a cold, mechanical, robot heart that favours hard facts over sentimentalism.  However, maybe s/the project/we/ for a slightly more human touch? :-)

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: I was trying to generalise for a concise name that could work for all examples. Hence employing some slightly fuzzy logic so that the ID in question may not be the absolute TA but is certainly an identifiable step towards it. DICE can be a pain here given both CA and self signed forms existing in close harmony. If we look at the purpose of this ID as being for modelling purposes then can we let this ride or do you think the lack of precision will distract?

SimonFrost-Arm:
> However, maybe s/the project/we/ for a slightly more human touch? :-)
I had some more personal wording in a couple of sections and took it out to be more consistent, maybe I corrected the wrong half?

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: I think the trusted epithet is implicit. How about 'first boot boot component'?

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: I had something along those lines and linked it to arbitrary queries for the 'other' section but then removed it as it sounded too vague and didn't influence the data model.

docs/device-and-supply-chain-modelling.md
thomas-fossati: the thing is "trust anchor" is an already taken term (e.g., see RFC4949), so reusing it with special meaning is bound to create dissonance in those that are already familiar with it.

docs/device-and-supply-chain-modelling.md
thomas-fossati: What I'm saying is that we might want to avoid special-casing this and treat version (or other more or less complex SW identifiers) as reference values in the same class as measurements.

docs/device-and-supply-chain-modelling.md
thomas-fossati: WFM -- I like the sound it makes: booooot boooooooot! :-)

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: OK, I'll ponder some text restoring some reference to RVs. I really wanted just to talk about the modelling required for measurements etc here & only included a reference to version roll up to respect that subset.

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: I'm not trying to redefine trust anchor, just define the use of a TA in this context. I still think using TA-ID will work for the purpose of relating to the generalised modelling. I've pushed a change with a more general description.

thomas-fossati (approved):
LvGTM and thanks for the changes!

docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,78 @@
+# Device and Supply Chain modelling
+
+## Intro
```
yogeshbdeshpande, line 3: Introduction would be better here than Intro.

docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Device evidence classification
+The evidence within an attestation report can be categorised into the following classes. These classes are not an exhaustive set of all types of evidence that may be found within a report but are those necessary for correlating evidence with endorsement / reference value queries.
+
+### Trust Anchor
+All devices are ASSUMED to contain a Trust Anchor (TA) or a descendant from a Trust Anchor. This is normally provisioned into the device at manufacturing time but may also be provisioned to the device on enrolling into a management structure. The TA takes the form of a key or a seed from which an Attestation signing key can be derived. The attestation report will contain information, referred to here as a Trust Anchor ID (TA-ID), which a Verifier can use to locate an endorsement providing information on the public value of the relevant key to use to validate the signature on a report. The Trust Anchor may have unique-to-device or group scope (i.e. multiple devices contain the same trust anchor). The latter case has been used as one solution for privacy protection use cases.
```
yogeshbdeshpande, line 17: Minor comment: To improve readability consider splitting the sentence into two simple sentences:

referred to here as a Trust Anchor ID (TA-ID), which a Verifier can use to locate an endorsement. This endorsement provides information on the public part of the relevant key, 
which is used to validate the signature of a report.

docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Hardware Group
+A device MAY contain a Hardware Group ID that is presented in attestation evidence. This is an identifier, provisioned during manufacturing, which can be used to classify the SoC used within the device and will be the same for any devices built using that SoC. Although termed as a hardware ID, in fact this identifier will classify the immutable Root of Trust, being the un-modifiable and un-measured first boot component as well as the specific SoC hardware within the SoC. The identifier is a public value and has no security value, but aids classification for a verifier. If no Hardware Group ID is included within the attestation evidence then an equivalent value can be provisioned into the verifier as a correlation to the Trust Anchor ID.
```
yogeshbdeshpande, line 20: The last line is particularly not clear to me?
If no Hardware Group ID is included within the attestation evidence then an equivalent value can be provisioned into the verifier as a correlation to the Trust Anchor ID.

May be please clarify the usage ? What added value does the "equivalent provides"?

docs/device-and-supply-chain-modelling.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Other Evidence
+Other evidence may be contained within the Attestation Report that does not fall into the above categories. This evidence may need to be referenced as part of an evidence appraisal policy, but no additional classification for a group of evidence is known at the moment to affect the modelling of Endorsements for the project.
+
+Note: attestation reports may contain additional complexities to modelling evidence, such as having multiple attesters within a device or receiving an encrypted report. However, it is expected that these can be handled as precursors to report verification by deconstruction into core parts in a pre-verification pass.
+
+Diagram: relevant groupings for evidence
+
+![Groupings](http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/veraison/veraison/main/docs/diags/tokens-model.puml)
```
yogeshbdeshpande, line 35: Somehow I was unable to view the links when using the Rich format?

yogeshbdeshpande (approved):
LGTM, left few minor comments but nothing major as such!

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: Fixed

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: Fixed

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: This refers to an identifier value to be used in place of a group claim in evidence.  Tweaked wording in latest push.

docs/device-and-supply-chain-modelling.md
SimonFrost-Arm: Odd, will take a look.

thomas-fossati: This is expected: the links point to the "future" diagrams, i.e., those that will appear on the main branch after merge.

PR #70, opened 2021-09-28
Introduce Provisioning Architecture Diagram

yogeshbdeshpande:


thomas-fossati:
I am not sure I understand this PR: Is this supposed to be merged without accompanying text or being referenced from existing documentation?

yogeshbdeshpande:
> I am not sure I understand this PR: Is this supposed to be merged without accompanying text or being referenced from existing documentation?

This is just a place holder so that we do not lose this. Yes eventually this will be embedded in a suitable text to be referred!

thomas-fossati:
> > I am not sure I understand this PR: Is this supposed to be merged without accompanying text or being referenced from existing documentation?
> 
> This is just a place holder so that we do not lose this. Yes eventually this will be embedded in a suitable text to be referred!

OK, let's track that future activity in a task though, otherwise in a couple of weeks this ends up being a mystery object :-)

yogeshbdeshpande:
> > > I am not sure I understand this PR: Is this supposed to be merged without accompanying text or being referenced from existing documentation?
> > 
> > 
> > This is just a place holder so that we do not lose this. Yes eventually this will be embedded in a suitable text to be referred!
> 
> OK, let's track that future activity in a task though, otherwise in a couple of weeks this ends up being a mystery object :-)

Agree, tracked it via : https://github.com/veraison/veraison/issues/71

thomas-fossati (commented):
As discussed, let's have only one (editable PNG), not both.  Otherwise we can run into consistency troubles.

yogeshbdeshpande:
> As discussed, let's have only one (editable PNG), not both. Otherwise we can run into consistency troubles.

Agree, let me introduce this!

thomas-fossati (approved):
fantastic, thanks!

PR #74, opened 2021-10-06
Provisioning Interface for Endorsement Store

yogeshbdeshpande:
Initial Feedback is Welcome!


endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
 }
+message SwComponentID {
+    common.TokenFormat type = 1;
+    google.protobuf.Struct id = 2;
+}
+// SwComponentInfo defines the structure of a single Software Component 
+// identified by a SoftwareComponent Identifer.
+message SwComponentInfo {
+    SwComponentID comp_id = 1;
+    google.protobuf.Struct comp_meta = 2;
```
thomas-fossati, line 54: we can simply call these `id` and `meta`

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+    SwComponentID comp_id = 1;
+    google.protobuf.Struct comp_meta = 2; 
+}
+
+message AddSwComponentArgs {
+    repeated SwComponentInfo info = 1;
+}
+
+message GetSwComponentArgs {
+    string software_comp_id = 1;
```
thomas-fossati, line 62: we can simply call this `id`

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+}
+
+message GetSwComponentRsp {
+    SwComponentInfo sw_component = 1;
+    int32 error_value = 2;
+    string error_detail = 3;
+}
+
+enum TaType {
+    UNSPEICIFIED = 0;
```
thomas-fossati, line 72: typo: `UNSPECIFIED`

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+    KEY = 1;
```
thomas-fossati, line 73: "KEY" is a tad too generic, unless we rely on the PEM encoded value to also carry type information.  But if the latter, we'd maybe want to drop TaType altogether.

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+    CERTIFICATE = 2;
+    INVALID = 3;
```
thomas-fossati, line 75: if we want this to be extensible we should move INVALID to the end of the type range

endorsement/storeapi.proto
yogeshbdeshpande: There is already an "id" inside the SwComponentID, hence this way of representation!

endorsement/storeapi.proto
yogeshbdeshpande: Good catch!

endorsement/storeapi.proto
yogeshbdeshpande: See https://github.com/veraison/veraison/issues/75 raised for this item

thomas-fossati: ok, one thing though: CERTIFICATE should probably be CERTCHAIN

endorsement/storeapi.proto
```diff
@@ -59,7 +59,7 @@ message AddSwComponentArgs {
 }
 
 message GetSwComponentArgs {
-    string software_comp_id = 1;
+    string id = 1;
```
thomas-fossati, line 62: this should be SwComponentID

yogeshbdeshpande: Why?

endorsement/storeapi.proto
yogeshbdeshpande: Yes, I noticed it already, will correct it!

endorsement/storeapi.proto
thomas-fossati: the id format is structured

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+    SwComponentID comp_id = 1;
+    google.protobuf.Struct comp_meta = 2; 
+}
+
+message AddSwComponentArgs {
+    repeated SwComponentInfo info = 1;
+}
+
+message GetSwComponentArgs {
+    SwComponentID id = 1;
```
thomas-fossati, line 62: In a previous comment you said _"There is already an "id" inside the SwComponentID, hence this way of representation!"_ which seems to apply here as well, right?  Should we call this `comp_id` as we do in `SwComponentInfo`?

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+}
+
+message GetSwComponentRsp {
+    SwComponentInfo sw_component = 1;
```
thomas-fossati, line 66: this doesn't look entirely right.  ISTM we should return meta and leave id out?

endorsement/storeapi.proto
thomas-fossati: Unresolving because I think CERTCHAIN is important.  If you don't think so please explain why.

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+
+enum TaType {
+    UNSPECIFIED = 0;
+    KEY = 1;
+    CERTIFICATE = 2;
+    INVALID = 255;
+}
+
+message TrustAnchorID {
+    common.TokenFormat type = 1;
```
thomas-fossati, line 79: We should be concerned with endorsement format rather than the evidence format.

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+    // Trust Anchor ID assumes a form based on the type of TokenFormat
+    google.protobuf.Struct id = 2;
+}
+
+message TrustAnchorValue {
+    TaType type = 1;
+
+    // The value is the actual trust anchor. It is a map with
+    // key-value pairs of TA. What part comprise of a trust anchor value 
+    // is defined by the type.
```
thomas-fossati, line 90: this comment is not very clear, I suggest some polishing, or even removing since it's pretty clear how this is supposed to work.

endorsement/storeapi.proto
```diff
@@ -43,6 +43,76 @@ message OpenArgs {
...
+    google.protobuf.Struct value = 2;
+}
+
+// Trust Anchor Info holds information about a Trust Anchor
+// A Trust Anchor is uniquely identified by a Trust Anchor ID, while
+// the actual Trust Anchor is conveyed via TrustAnchorValue
```
thomas-fossati, line 96: I don't think this comment is adding any useful information on top of what is already a very clear struct layout. Suggest removing it.

endorsement/storeapi.proto
```diff
@@ -53,5 +123,13 @@ service Fetcher {
     rpc GetEndorsements(GetEndorsementsArgs) returns (GetEndorsementsResponse);
 }
 
-
+// Provisioner is the Service end point that is used to provision
+// Software Components and Trust Anchors to and from
+// the Endorsement Store
```
thomas-fossati, line 128: "provision from" is not correct.  In general, I don't think these comments are very helpful.  I suggest we just drop them.

endorsement/storeapi.proto
yogeshbdeshpande: Good Point, I can change this to be consistent!

endorsement/storeapi.proto
thomas-fossati: >  It is effectively a single unit!

why?  Sorry I don't get this.

endorsement/storeapi.proto
```diff
@@ -43,6 +63,62 @@ message OpenArgs {
...
+
+message AddTrustAnchorArgs {
+    repeated TrustAnchorInfo trust_anchors = 1;
+}
+
+message GetTrustAnchorArgs {
+    TrustAnchorID trust_id = 1;
+}
+
+message GetTrustAnchorRsp {
```
thomas-fossati, line 117: align typographic conventions with existing GetEndorsements: `GetTrustAnchorResponse`

endorsement/storeapi.proto
```diff
@@ -43,6 +63,62 @@ message OpenArgs {
...
+
+message AddSwComponentArgs {
+    repeated SwComponentInfo info = 1;
+}
+
+message GetSwComponentArgs {
+    SwComponentID comp_id = 1;
+}
+
+message GetSwComponentRsp {
```
thomas-fossati, line 85: let's align typo conventions with existing GetEndorsements: `GetSwComponentResponse`

endorsement/storeapi.proto
```diff
@@ -43,6 +63,62 @@ message OpenArgs {
...
+}
+
+message GetTrustAnchorArgs {
+    TrustAnchorID trust_id = 1;
+}
+
+message GetTrustAnchorRsp {
+    TrustAnchorValue value = 1;
+    int32 error_value = 2;
+    string error_detail = 3;
```
thomas-fossati, line 120: these two should go first (renumber)

endorsement/storeapi.proto
```diff
@@ -43,6 +63,62 @@ message OpenArgs {
...
+}
+
+message GetSwComponentArgs {
+    SwComponentID comp_id = 1;
+}
+
+message GetSwComponentRsp {
+    google.protobuf.Struct comp_meta = 1;
+    int32 error_value = 2;
+    string error_detail = 3;
```
thomas-fossati, line 88: these two should go first

endorsement/storeapi.proto
```diff
@@ -43,6 +63,62 @@ message OpenArgs {
...
+    google.protobuf.Struct value = 2;
+}
+
+message TrustAnchorInfo {
+    TrustAnchorID id = 1;
+    TrustAnchorValue value = 2;
+}
+
+message AddTrustAnchorArgs {
+    repeated TrustAnchorInfo trust_anchors = 1;
```
thomas-fossati, line 110: What does it mean to Add multiple TAs in one go?  I am not sure I understand the use case for this.

thomas-fossati (commented):
I have left a number of mostly editorial comments to align with Sergei's conventions.

Two high level comments:

* Do we want error_value values to be defined as part of the interface?  Or are they defined on top of the transport containers in a wrapping library?

* We should probably have Meta added to TA and SwComp to deal with Information that isn’t part of the value per se but can be useful for other purposes, e.g.: defining update/patch relations, invalidating entries (e.g., known bad values, revoked keys)

endorsement/storeapi.proto
yogeshbdeshpande: ok

endorsement/storeapi.proto
yogeshbdeshpande: Agree, but what are the chances of failure use case you depicted and losing the entire payload versus the benefit of adding in single message.
Protobuf is precisely very good at such serialization! Hence I see a value in adding the interface like this!

thomas-fossati: > Processing cost: Its fairly obvious that if the information is present to the sender, it is much more efficient to send it in one bundle against repeated messages of the same nature over the wire!

of course, at a micro-optimisation level there is no question.  But how that impacts on the overall processing is impossible to estimate at this point in time.  IOW, how do we know this is going to be a bottleneck?  And the answer is until we measure, we can't know.

> Interface correctness and ergonomics are not compromised laying out interface in this manner!

They are, as I have explained above.

> I am opposed to adding an Identical message : _AddManyTrustAnchors_ interface, as it duplicates the existing interface with no added value. if one trust anchor is needed, the count can be one, if more than one then the count can be much more!

Ditto.

> The same principle has been followed in Software Components also!

I don't think the analogy is not correct, because SwComponents (AFAIU) are bundled around the same "update" block, so it makes sense to cluster them and make them share fate within the same transaction.

> The interface is adding flexibility to set multiple items of identical type in the interface!

See above.

thomas-fossati:
> > * we want error_value values to be defined as part of the interface?  Or are they defined on top of the transport containers in a wrapping library
> 
> This depends on the what we define as Error Value. If this is per message specific, with value tied to semantic then this is the right place for it to be! It is more coupled to the message specifics!

As discussed, I think this may also be approached in a more generic way, by establishing broad error categories that are valid for each RPC endpoint rather than being per-endpoint.  The simplest would be only have a binary signal, i.e., to transform `int error_value` into `bool success`, and let `string error_details` encode the human-readable description of what happened.
```
Error {
  bool success,
  string details,
}
```
A slightly richer solution could allow distinguishing errors that depend on client input from those that are purely server-side problems (e.g., temporary unavailability of a needed resource).  Note that this could be built on top of the previous by adding a `Category` (maybe using distinct ranges) enum :
```
Error {
  bool success,
  string details,
  Category {
    CLIENT,
    SERVER_TEMPORARY,
    SERVER_PERMANENT,
    etc.
  } category,
}
```

endorsement/storeapi.proto
```diff
@@ -42,7 +42,7 @@ message GetEndorsementsArgs{
 }
 
 message Response {
-    int32 error_value = 1;
+    bool result = 1;
     string error_detail = 2;
```
thomas-fossati: I wonder if we should define a top-level Status message that we can extend independently

endorsement/storeapi.proto
```diff
@@ -44,6 +63,58 @@ message OpenArgs {
...
 }
 
+message SwComponentID {
+    AttestationFormat type = 1;
+    google.protobuf.Struct parts = 2;
+}
+
+message SwComponent {
+    SwComponentID id = 1;
+    google.protobuf.Struct meta = 2;
```
thomas-fossati, line 73: I thought we agreed on having meta in addition to the (id, value), similar to TrustAnchor.

endorsement/storeapi.proto
```diff
@@ -5,10 +5,30 @@ option go_package = "github.com/veraison/endorsement";
...
+	// PSA-IOT is the PSA attestation format (based on:
+	// https://developer.arm.com/architectures/architecture-security-features/platform-security)
+	PSA_IOT = 1;
+
+	// DICE is the attestation format based on the TCG DICE specification
+	// https://trustedcomputinggroup.org/wp-content/uploads/TCG_DICE_Attestation_Architecture_r22_02dec2020.pdf
+	DICE = 2;
+}
+
+enum TaType {
```
setrofim, line 10: I would spell the name out fully -- ``TrustAnchorType``; it's not that long, and is a lot clearer at a glance.

common/evidence.go
```diff
@@ -14,11 +14,11 @@ type EvidenceContext struct {
 	// extracted. This is used to specify how the Evidence structure should
 	// be interpreted, and to identify which endorsements will be necessary
 	// for verification.
-	Format TokenFormat `json:"token_format"`
+	Format AttestationFormat `json:"token_format"`
```
setrofim, line 17: Should rename JSON field as well.

thomas-fossati (approved):
Thanks!

PR #79, opened 2021-11-03
DICE documentation

thomas-fossati:
For a preview: https://github.com/veraison/veraison/blob/dice-docs/docs/dice.md


docs/dice.md
```diff
@@ -0,0 +1,170 @@
...
+
+### TCG DICE Formats
+
+```asn.1
+tcg OBJECT IDENTIFIER ::= { 2 23 133 }
+
+tcg-dice OBJECT IDENTIFIER ::= { tcg platformClass(5) 4 }
+```
+
+#### TCB Info
```
SimonFrost-Arm, line 65: Isn't this the TCI? Definition above but not otherwise mentioned. Adjust to correct terminology

docs/dice.md
```diff
@@ -0,0 +1,170 @@
...
+> NOTE: This mode shifts all the costs on the verification infra.  It is an operational nightmare with a vast attack surface and very limited scalability.  It should be deemed acceptable only when the cost of a key compromise / disclosure is (very) low, and the device fleet is roughly static with very infrequent software updates.
+
+### Verification
+
+We expect that, typically, DICE evidence submitted for verification is not bound to an explicit challenge-response session.  This is likely to require defining a new Veraison API alongside the existing [challenge-response](api/challenge-response/README.md) endpoint.
+
+Given the fact that TCG DICE has a few loose ends, mainly in terms of evidence encoding formats and the nature of cryptographic identities, Veraison will have to be as flexible as possible while clearly documenting the supported feature set.
+
+Open-DICE defines a stricter (and smaller) set of requirements and therefore it'd be probably easier to implement fully.
+
```
SimonFrost-Arm, line 159: Is it worth noting here that currently neither spec discusses on-wire format / serialisation & that adds uncertainty to the verification process?

docs/dice.md
```diff
@@ -0,0 +1,170 @@
...
+> NOTE: At first it might seem like the scalability properties of DICE verification vis-a-vis software & configuration updates of TCB components are terrible.  The verifier would have to be re-provisioned with all the updated identities of all the devices of the same "class" if a bit was flipped in their TCB.  Luckily, the situation is not so gloomy: as long as the lowest layer in the DICE chain is stable (as in "never updated") and the DICE compound identity is provided as a certification chain rooted either in the lowest layer or in an external CA that issues the cert for the lowest layer, an identity update of the i-th layer can be re-computed dynamically by a verifier without need to further interact with the supply chain.  The key requirement is the stability of the so-called "Layer 0" / "First Mutable Code" (FMC), and therefore of its cryptographic identity.  The latter is a pretty important bit in DICE and it's given the proper name of "Device Id".  The Device Id key pair is associated with an IEEE 802.1AR IDevID cert.
```
SimonFrost-Arm, line 41: The verification process is more practical (accurate?) if the certs contain the Digest/Hash info for the current layer. These are currently marked as optional in both specs. Can we add some text that makes a recommendation for implementations to include these entries?

docs/dice.md
```diff
@@ -0,0 +1,170 @@
...
+  configurationHash       [2] EXPLICIT OCTET STRING OPTIONAL,
+  configurationDescriptor [3] EXPLICIT OCTET STRING OPTIONAL,
+  authorityHash           [4] EXPLICIT OCTET STRING OPTIONAL,
+  authorityDescriptor     [5] EXPLICIT OCTET STRING OPTIONAL,
+  mode                    [6] EXPLICIT Mode OPTIONAL,
+}
+```
+
+For further details regarding the Open DICE X.509 profile, see the [Certificate Details](https://pigweed.googlesource.com/open-dice/+/HEAD/docs/specification.md#certificate-details) section of the spec.
+
```
SimonFrost-Arm, line 131: Add a note to the effect that the openDICE preference is for the cbor encoding but that it will support mixed modes in different layers for legacy component integrations?

docs/dice.md
thomas-fossati:
```suggestion
> NOTE: At first it might seem like the scalability properties of DICE verification vis-a-vis software & configuration updates of TCB components are terrible.  The verifier would have to be re-provisioned with all the updated identities of all the devices of the same "class" if a bit was flipped in their TCB.  Luckily, the situation is not so gloomy: as long as the lowest layer in the DICE chain is stable (as in "never updated") and the DICE compound identity is provided as a certification chain rooted either in the lowest layer or in an external CA that issues the cert for the lowest layer, an identity update of the i-th layer can be re-computed dynamically by a verifier without need to further interact with the supply chain -- except for re-provisioning the verifier with the updated measurement.  The key requirement is the stability of the so-called "Layer 0" / "First Mutable Code" (FMC), and therefore of its cryptographic identity.  The latter is a pretty important bit in DICE and it's given the proper name of "Device Id".  The Device Id key pair is associated with an IEEE 802.1AR IDevID cert.
```

thomas-fossati: > The verification process is more practical (accurate?) if the certs contain the Digest/Hash info for the current layer. These are currently marked as optional in both specs. Can we add some text that makes a recommendation for implementations to include these entries?

I don't think we can make any recommendation from here.  We are just sort of passively documenting what happens.  What the text was missing was acknowledging the fact that the measurement (e.g., hash, version, etc.) is an integral part of the verification, not just the chaining of the certs.

docs/dice.md
```diff
@@ -0,0 +1,170 @@
+# DICE
+
+## Standard Documents
+
+Device Identifier Composition Engine (DICE) is an attestation standard developed by the Trusting Computing Group (TCG).  DICE builds on the device identity and device attestation concepts introduced in [Robust Internet-of-Things (RIoT)](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/06/RIoT20Paper-1.1-1.pdf) from Microsoft Research.
+
+The DICE specification is split across the following core documents:
+
+* [Hardware Requirements](https://trustedcomputinggroup.org/resource/hardware-requirements-for-a-device-identifier-composition-engine/), describing the DICE Root of Trust (RoT) and the construction of the (unique) Compound Device Identifier (CDI). 
+* [Layering Architecture](https://trustedcomputinggroup.org/resource/dice-attestation-architecture/), describing the construction of a device's layered identity rooted in the CDI and obtained via the recursive invocation of a simple primitive based on measurements and per-device secret seeding.
```
thomas-fossati, line 10:
```suggestion
* [Layering Architecture](https://trustedcomputinggroup.org/resource/dice-layering-architecture/), describing the construction of a device's layered identity rooted in the CDI and obtained via the recursive invocation of a simple primitive based on measurements and per-device secret seeding.
```

docs/dice.md
thomas-fossati: The TCI  is the component identifier.  It can be derived by hashing a bunch of measurements associated with a component (e.g., hash of the firmware, model, version, svn, etc.) into one single scalar.   The TCB Info gathers a standard set of features associated with the components into a multi-dimensional vector.

docs/dice.md
thomas-fossati: Are we sure Open-DICE prefers CBOR?  Anyway, you are right we should say something about the CBOR format and the mixed mode.

docs/dice.md
thomas-fossati: Are you thinking of the lack of media types for transporting DICE material?

docs/dice.md
thomas-fossati:
```suggestion

Open DICE also allows encoding UDS and CDI certificates in [CBOR Web Tokens (CWT)](https://pigweed.googlesource.com/open-dice/+/HEAD/docs/specification.md#CBOR-UDS-Certificates) similar to [EAT](https://datatracker.ietf.org/doc/draft-ietf-rats-eat/).   It also allows for hybrid cert chains, with co-existing X.509 and CWT certs in the same logical chain, to support legacy component integrations.
```

docs/dice.md
SimonFrost-Arm: Ah, got it. Can we clarify please otherwise the TCI defn is rather left hanging.

docs/dice.md
SimonFrost-Arm: That's a specific, but I was generally thinking of this as a challenge for verification that the specs don't address, hence leaving the prospect for an (open/TCG)) DICE deployment based on Veraison as potentially needing to add some custom handling at the front end rather than being a full reference

docs/dice.md
SimonFrost-Arm: OK, good point, just my wishful thinking!

docs/dice.md
thomas-fossati: sure, I added an entry in the glossary

docs/dice.md
thomas-fossati: The open-DICE spec is much more crisp / easy to implement than TCG DICE.  I think the only real problem there is the potential for hybrid CWT/x.509 chains.

TCG DICE on the contrary will necessarily need some hammering to be fully supported.  But this point  is articulated already in L156.  Or do you think that is not sufficient?

SimonFrost-Arm: Thinking about it some more, that probably covers it, I may be just being over sensitive to what a consuming parties expectations might be, but we have time to accommodate such conversations.

SimonFrost-Arm (approved):

PR #86, opened 2021-11-16
Modify color codes to create more meaningful diagrams

yogeshbdeshpande:


SimonFrost-Arm (approved):
Thanks, rhat fixes the accidental visual link I was concerned about

PR #90, opened 2021-11-23
Rename "store" command to "vts"

setrofim:
It has become clear that what we previously referred to as the
"endorsement store node" would have a wider set of responsibilities than
just handling endorsements. As such, that name was misleading and a new
name -- Veraison Trusted Services, or VTS (pronounced "vitis") -- was
chosen as more representative of the node's purpose.


thomas-fossati:
cool! Looks like there's a small thing missing:
```
diff --git a/cmd/Makefile b/cmd/Makefile
index 6dad9f5..e272aff 100644
--- a/cmd/Makefile
+++ b/cmd/Makefile
@@ -3,6 +3,6 @@

 #SUBDIR := endorsements
 #SUBDIR += policy
-SUBDIR += store
+SUBDIR += vts

include ../mk/subdir.mk
```

setrofim:
Yup, d'oh. It's now fixed.

setrofim:
*sigh* also missed the .mod... OK, 3rd time lucky :crossed_fingers:

thomas-fossati (approved):

PR #106, opened 2021-12-09
key-val store

### LINKED ISSUE #97: Feature: key-val store
Implement Veraison storage layer.
Initially this is a simple key-val store.
There must exist an interface that decouples the calling code from the specific DB technology used to implement it.

thomas-fossati:
interface and two implementations:
* generic SQL backend
* in-memory

Fixes #97

### DIFF
--- a/kvstore/README.md
+++ b/kvstore/README.md
@@ -1,0 +1,77 @@
+# KV Store
+
+The key-value (KV) store is Veraison storage layer.  It is used for both endorsements and trust anchors.
+
+It is intentionally "dumb": we assume that the filtering smarts are provided by the plugin.
+
+The key is a string synthesised deterministically from a structured endorsement / trustanchor identifier.  It is formatted according to a custom URI format -- [see below](#uri-format)).
+
+The value is a string containing the endorsement or trust anchor data in JSON.  This data varies depending on the attestation format.
+
+The `KVStore` interface defines the required methods for storing, fetching and deleting KV objects.  Note that there is no method for patching data in place.  Methods for initialising and orderly terminating the underlying DB are also provided.
+
+This package contains two implementations of the `KVStore`:
+
+1. `SQL`, supporting different SQL engines (e.g., SQLite, PostgreSQL, etc. -- [see below](#sql-drivers)),
+1. `Memory`, a thread-safe in-memory associative array intended for testing.
+
+## SQL drivers
+
+To use a SQL backend the calling code needs to (anonymously) import the supporting driver.
+
+For example, to use PostgreSQL:
+```go
+import _ "github.com/lib/pq"
+```
+Instead, to use SQLite:
+```go
+import _ "github.com/mattn/go-sqlite3"
+```
+
+## SQL schemas
+
+```sql
+CREATE TABLE endorsement (
+  key text PRIMARY KEY,
+  val text NOT NULL
+);
+
+CREATE TABLE trustanchor (
+  key text PRIMARY KEY,
+  val text NOT NULL
+);
+```
+
+## URI format
+
+```abnf
+scheme ":" authority path-absolute
+```
+
+where:
+
+* `scheme` encodes the attestation format (e.g., "psa", "tcg-dice",
+"tpm-enacttrust", "open-dice", "tcg-tpm", etc.)
+* `authority` encodes the tenant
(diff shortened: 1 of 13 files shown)


kvstore/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+where:
+
+* `scheme` encodes the attestation format (e.g., "psa", "tcg-dice",
+"tpm-enacttrust", "open-dice", "tcg-tpm", etc.)
+* `authority` encodes the tenant
+* `path-absolute` encodes the parts of the key, identified positionally.  Missing optional parts are encoded as empty path segments.
+
+Attestation techology specific code (i.e., plugins) must provide their own syntesis functions.
```
yogeshbdeshpande, line 58:
```suggestion
Attestation technology specific code (i.e., plugins) must provide their own synthesis functions.
```

kvstore/kvstore.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2021 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package kvstore
+
+type KVStore interface {
+	Init(cfg Config) error
+	Term() error
```
yogeshbdeshpande, line 7:
```suggestion
	Close() error
```

kvstore/memory_test.go
```diff
@@ -0,0 +1,234 @@
...
+		"store-type": "some-random-type",
+	}
+
+	expectedErr := `invalid "store-type": unknown type "some-random-type"`
+
+	err := s.Init(cfg)
+	assert.EqualError(t, err, expectedErr)
+}
+
+func TestMemoryStore_Init_bad_store_type(t *testing.T) {
```
yogeshbdeshpande, line 58: Tests on Line 32, 45 and 58 can be potentially combined into a single table driven test.

thomas-fossati: True, and others as well in this same lot could be made table-driven, I think.  Question: is it necessary?

kvstore/kvstore.go
thomas-fossati: done in 0417174

kvstore/sql_test.go
```diff
@@ -0,0 +1,298 @@
...
+	e.WithArgs(testKey)
+	e.WillReturnRows()
+
+	err = s.Del(testKey)
+	assert.NoError(t, err)
+
+	if err := mock.ExpectationsWereMet(); err != nil {
+		t.Errorf("unfulfilled expectations: %s", err)
+	}
+}
```
yogeshbdeshpande, line 298: I do not see any test which tests TestSQLStore_Init_Close_cycle_ok ?
I see close() executed naturally, I understand, but where is s.Init(cfg) +ive case?

yogeshbdeshpande (commented):
Have completed my review!

kvstore/memory_test.go
yogeshbdeshpande: Well, only benefit I see is the code compactness and logical coherence! Reader sees all negative tests in one place.

I leave this up to you on your best judgement!

kvstore/sql_test.go
thomas-fossati: oops, I forgot.  Will add.

kvstore/memory_test.go
thomas-fossati: On compactness I agree, logical coherence not so much.  Anyway, if you really prefer this way, I'll do it.

kvstore/sql_test.go
thomas-fossati: Fixed in 0dfbb0e

kvstore/memory_test.go
thomas-fossati: Addressed in 0dfbb0e

kvstore/sql_test.go
```diff
@@ -71,6 +71,22 @@ func TestSQL_Init_db_open_unknown_driver_postgres(t *testing.T) {
 	assert.EqualError(t, err, expectedErr)
 }
 
+func TestSQL_Init_Close_cycle_ok(t *testing.T) {
+	s := Memory{}
```
thomas-fossati, line 75: arg!

thomas-fossati: it turns out we cannot run this test because it needs a SQL driver to work (see README.md)

yogeshbdeshpande (approved):

Issue #122, opened 2022-02-22
Store API: remove distinction between id and value in the  endorsements


thomas-fossati:
We need to remove the (artificial) distinction between id and value in the endorsement payloads.  At that level, everything is just an endorsement: the lookup keys will be dynamically synthesised by the appropriate plugin depending on the attestation format and the set of access queries.

PR #129, opened 2022-03-14
Endorsement refactor

yogeshbdeshpande:
First Phase of Endorsement Refactor


common/vts.proto
```diff
@@ -15,6 +19,122 @@ message CloseArgs {
...
+}
+
+message GetEndorsementsResponse {
+    Status status = 1;
+    string details = 2;
+    TrustVector trust_vector = 3;
+    google.protobuf.Struct endorsements = 4;
+}
+
+
```
thomas-fossati, line 65:
```suggestion
```

mk/test.mk
```diff
@@ -34,11 +34,12 @@ test-hook-pre:
 realtest: _mocks ; go test $(TEST_ARGS) $(GOPKG)
 .PHONY: realtest
 
+COPYRIGHT_FLAGS :=
```
thomas-fossati, line 37: looked correct to me

thomas-fossati (commented):
LGTM

common/vts.proto
```diff
@@ -15,6 +19,122 @@ message CloseArgs {
...
+}
+
+enum TaType {
+    TA_UNSPECIFIED = 0;
+    TA_RAWPUBLICKEY = 1;
+    // Refer https://github.com/veraison/veraison/issues/76, to introduce certs here!
+    TA_INVALID = 255;
+}
+
+message EndorsementID {
```
setrofim, line 39: Shouldn't this be URI as defined here; https://github.com/veraison/veraison/blob/main/kvstore/README.md?

common/vts.proto
```diff
@@ -15,6 +19,122 @@ message CloseArgs {
...
+message GetSwComponentRequest {
+    SwComponentID id = 1;
+}
+
+message GetSwComponentResponse {
+    Status status = 1;
+    google.protobuf.Struct attributes = 2;
+}
+
+message TrustAnchID {
```
setrofim, line 93: This naming seems inconsistent with `TaType` above and `TrustAnchorValue` below.  I suggest aligning the three on one of

- `TAType`, `TAID`, `TAValue`
- `TrustAnchType`, `TrustAnchID`, `TrustAnchValue`
- `TrustAnchorType`, `TrustAnchorID`, `TrustAnchorValue`

common/vts.proto
```diff
@@ -15,6 +19,122 @@ message CloseArgs {
...
+}
+
+message OpenArgs {
+}
+
+message OpenResponse {
+    Status status = 1;
+}
+
+enum TaType {
```
setrofim, line 32: Should be `TAType` to be consistent with golang naming conventions?

common/vts.proto
```diff
@@ -15,6 +19,122 @@ message CloseArgs {
...
+
+message GetTrustAnchorResponse {
+    Status status = 1;
+    TrustAnchorValue value = 2;
+}
+
+// Provisioner is the Service end point that is used to store
+// and retrieve Software Components and Trust Anchors to and from
+// the Endorsement Store
+service Provisioner {
```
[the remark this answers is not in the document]
yogeshbdeshpande: @setrofim : Yes, we have discussed this briefly! Store will be removed in the next phase and we will have a Logical VTS Endpoint, abstraction which takes on a Verifier Service, Provisioner Service and Common Service. Will be done in net phase!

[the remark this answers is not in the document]
setrofim: What was the rationale for separating the Verification and Provisioning services (and thus necessitating a Common service, thus complicating the use of VTS). Why not just have `VTS` service that supports the verification and provisioning APIs (plus whatever other interfaces we might need in the future)?

(IIRC, our original discussion that resulted in separating the provisioning and verification interfaces into separate entities was when the provisioner and the verifier were talking to the endorsemetn store directily; now that we have the VTS that mediates access to it, I'm not sure the old arguments still hold).

common/vts.proto
thomas-fossati: This is bound to go in https://github.com/veraison/veraison/issues/122

common/vts.proto
thomas-fossati: SGTM

common/vts.proto
setrofim: Why do we need two `TrustAnchor`s? shouldn't the other one be removed?

thomas-fossati: > Why do we need two `TrustAnchor`s? shouldn't the other one be removed?

yep!  I think Yogesh wasn't sure and wanted to discuss the need of the other with you before dropping it.  If you are OK with `rm trustanchor.go` let's do it :-)

common/vts.proto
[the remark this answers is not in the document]
setrofim: It seems like we adding actual complexity (having to deal with multiple interfaces -- at least two: common, and provisioner/verifier) for a hypothetical benefit (being able to split VTS across multiple nodes). You're also making the assumption that if we were going to split VTS, it would be along provisioner/verifier lines. I don't see why that would necessarily be the case. 

I think it might be better to err on the side of simplicity. If/when it turns out we want to split VTS services, we can always add the necessary interfaces later -- note that since we taking about interfaces here, it should be fairly straight forward. We can always maintain backwards compatibility by providing a "shim" VTS node implementation that forward requests to individual service nodes.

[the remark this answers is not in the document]
thomas-fossati: Another (more concrete) reason not to merge the services is that the client-side (i.e., the provisioning and verification frontends) would otherwise need to provide implementations for all the VTS RPCs, when all they need is to interface only the relevant subset (e.g., in the case of provisioning implementing the `common.ProvisionerClient` interface).

common/vts.proto
thomas-fossati: > @setrofim @thomas-fossati , Yes, I am aware that we need to remove it. It will be done in Phase 2. I just want to avoid a bulk PR and would like to go small steps at a time!

I think if we all agree that bit is not needed we might as well drop it in this same PR.

common/vts.proto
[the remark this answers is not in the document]
setrofim: We only need one client side implementation (at least, for a given channel). Provisioner and Verifier will then both use  the VTS client. There is no reason either the Provisioner or the Verifier need to worry about RPC.

common/vts.proto
setrofim: > If you are OK with rm trustanchor.go let's do it :-)
SGTM

yogeshbdeshpande: ok, I am changing it now, as we speak!

common/vts.proto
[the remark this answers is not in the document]
thomas-fossati: The problem I see with this approach is that a VTSClient instantiated on the provisioning pipeline will be able to request verification-specific services.  This looks problematic least-privilege-wise.


But one thing I am not fully grasping (which may be worth discussing IRL) is the added complexity you mention.  For example, these are the [interface](https://github.com/veraison/veraison/blob/provisioning-part2/provisioning/storeclient/istoreclient.go) and the [implementation](https://github.com/veraison/veraison/blob/provisioning-part2/provisioning/storeclient/storeclient_grpc.go) of the VTS client on the provisioning side (they currently sit in the `provisioning-part2` branch): they don't look especially complex to me.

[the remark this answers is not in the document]
yogeshbdeshpande: This will be resolved during Phase-2 of the Refactor.

setrofim (approved):
LGTM (modulo the client issue to be handled in phase 2)

common/vts.proto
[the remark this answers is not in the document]
yogeshbdeshpande: Having read the discussion more deeply and thinking a bit more, I think, I am aligning my thought process with @setrofim.
Simple single VTS Interface to VTS Clients, with an Access Layer to be built on top of it, giving specific API Access of VTS to specific clients can lead to a simple solution. Looking forward to discuss this aspect during the meeting!

thomas-fossati (approved):

PR #132, opened 2022-03-18
Merge the Provisioner service into VTS

thomas-fossati:
This change merges the Provisioner service into VTS.  It also removes
the Store service, and drops Init and Close from the set of VTS's
exposed RPC endpoints.


setrofim (approved):
LGTM!

PR #134, opened 2022-03-30
Integrate the provisioning path into VTS

### LINKED ISSUE #92: Feature: Provisioning Pipeline
Implement Provisioning Pipeline for Endorsement Provisioning for Veraison

thomas-fossati:
The main deliverable of this changeset is the addition of
provisioning-side API logics to the VTS plugins (PSA and TPM EnactTrust).

It also contains a number of significant changes to the KV store which
are listed below:

* Allow multiple values per key
* Rework kvstore configuration logics to integrate it better into the
  new Veraison config
* Rename KVStore interface type to IKVStore
* Use Exec for INSERTs and DELETEs in the SQL backend
* Add a factory method to instantiate IKVStore's from config

Signed-off-by: Thomas Fossati <thomas.fossati@arm.com>

Fixes #92


trustedservices/rpc.go
```diff
@@ -5,19 +5,24 @@ package trustedservices
...
+	if store == nil {
+		return nil, errors.New("param store initialization failed")
+	}
 	err := PopulateRPCServerParams(store)
 	return store, err
 }
 
 func PopulateRPCServerParams(store *common.ParamStore) error {
 	return store.AddParamDefinitions(map[string]*common.ParamDescription{
+		// NOTE(tho) should we use Path as map key?  (It'd make debug messages clearer)
```
setrofim, line 25: This would tie a param to a specific path within a config file. I explicitly tired to avoid that (hence referring to it by a name, rather than path).  A config file is tied to a particular deployment/configuration of Veraison components; e.g., I would expect a config file for a single-binary build of integrated Veraison services to look different from the config file for VTS node.

This would also make it easier to embed Veraison config as a sub-tree in a larger service config file, or even to re-use setting from an existing config file to initialize Veraison components.

kvstore/ikvstore.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2021-2022 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package kvstore
+
+type IKVStore interface {
+	Init(cfg Config) error
+	Close() error
+	Get(key string) ([]string, error)
+	Set(key string, val string) error
```
yogeshbdeshpande, line 9: Since now for a key, multiple values can exist, should the interface for Set should also reflect the same?

yogeshbdeshpande (commented):
Just started  with few comments I will review it more thoroughly later today!

cmd/vts/main.go
```diff
@@ -11,13 +11,15 @@ import (
...
+
+	_ "github.com/mattn/go-sqlite3"
 )
 
 // TODO this a very minimal "frontend" implementation.
 func main() {
 	configPaths := common.NewConfigPaths()
 
-	flag.Var(configPaths, "config", "Path to direcotory containing the config file(s).")
+	flag.Var(configPaths, "config", "Path to directory containing the config file(s).")
```
yogeshbdeshpande, line 22: Just a note, as doing formal thorough review. What does a //TO DO here reflect. Can we translate into a clear issue for a proper front end, if this is the ask here?

cmd/vts/test/config.yaml
```diff
@@ -1,10 +1,24 @@
...
+  locations:
+    - ../../plugins/bin/
+
 vts:
   port: 50051
+
 endorsement:
-    store_name: memory
+  kvstore_config:
+    type: endorsement
```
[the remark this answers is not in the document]
thomas-fossati: At the moment, yes.  That said, I don't think that the way the kvstore is supposed to "know" about the purpose of the underlying store is correct.  I have a couple of things that I want to modify there and one is that, which would remove the need for exposing `type`.

common/scheme_plugin.go
```diff
@@ -148,32 +206,76 @@ func (s *SchemeRPC) GetFormat() AttestationFormat {
...
+
+	err = s.client.Call("Plugin.SynthKeysFromSwComponent", args, &resp)
+	if err != nil {
+		return nil, fmt.Errorf("failed call to SynthKeysFromSwComponent: %w", err)
+	}
+
+	return resp, nil
+}
+
+func (s *SchemeRPC) SynthKeysFromTrustAnchor(tenantID string, ta *TrustAnchor) ([]string, error) {
```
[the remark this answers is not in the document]
thomas-fossati: > Shouldn't this be returning the same thing as `GetTrustAnchorID` below, with the only difference being the input it operates on?

Not really.  `GetTrustAnchorID` uses one specific name of the object it needs to retrieve, whereas the `SynthKeysFrom...` may potentially create different aliases for the same underlying object to be used in different queries, i.e., not just `GetTrustAnchorID` but also the other provisioning APIs which may have different access patterns.
 
> Also, for the sake of consistency in the APIs, we should also align on whether the `tenantID` is assumed to be part of a structure (as with `GetTrustAnchorID` below), or is always passed as the first argument.

True.

> (Note: I don't have a strong preference either way, and am happy for the tenant ID to be removed from the `AttestationToken` ).

I don't have either :-)

trustedservices/rpc.go
thomas-fossati: I understand the principle.  While fiddling with the config file, though, I was getting errors referring the symbolic name which is only available in the code and (we should assume) invisible to the end user.  And that usability-wise is not particularly friendly :-)

kvstore/ikvstore.go
thomas-fossati: You are right, this is another thing I want to change in kvstore (the other was mentioned in [this comment](https://github.com/veraison/veraison/pull/134#discussion_r838471280).  In fact, this should be `Append`, not `Set`.

cmd/vts/main.go
thomas-fossati: I think we can probably remove the `TODO` and leave the rest of the comment.  @setrofim ?

common/scheme_plugin.go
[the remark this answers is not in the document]
setrofim: > Not really. GetTrustAnchorID uses one specific name of the object it needs to retrieve, whereas the SynthKeysFrom... may potentially create different aliases for the same underlying object to be used in different queries, i.e., not just GetTrustAnchorID but also the other provisioning APIs which may have different access patterns.

Mm OK, I think I have a different intuition for what is implied by "key". I gather this intended to be used more like indexes or query parameters? Also, if they are intended to be used for different quires and in different contexts, how does one identify which query/context a particular result is to be used in? Is this intended to be implicit in the positional index of the value within the slice? If so, perhaps it would be clearer to make it explicit by making the different queries fields in a struct, or (if this is going to be scheme-specific) keys in a map?

kvstore/ikvstore.go
thomas-fossati: See https://github.com/veraison/veraison/issues/135

cmd/vts/test/config.yaml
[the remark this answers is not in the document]
thomas-fossati: See https://github.com/veraison/veraison/issues/135

trustedservices/rpc.go
setrofim: True, we should probably modify the error message to include the Path as well.

cmd/vts/main.go
setrofim: Yeah,  agreed -- probably just remove the `TODO` bit -- it's entended to reflect that the whole implementation is intended to be quick-and-dirty, rather than complete.

common/scheme.go
```diff
@@ -20,5 +20,8 @@ type IScheme interface {
 	GetFormat() AttestationFormat
 	GetTrustAnchorID(token *AttestationToken) (string, error)
 	ExtractEvidence(token *AttestationToken, trustAnchor string) (*ExtractedEvidence, error)
-	GetAttestation(ec *EvidenceContext, endorsements string) (*Attestation, error)
+	GetAttestation(ec *EvidenceContext, endorsements []string) (*Attestation, error)
```
yogeshbdeshpande, line 23: Noticed @thomas-fossati  has changed the Interface for GetAttestation as well. Does that impact Verifier code @setrofim ?

common/scheme_plugin.go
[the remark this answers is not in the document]
thomas-fossati: > [...] or (if this is going to be scheme-specific) keys in a map?

That's a really intriguing idea, I mean this notion of storing the key purpose (i.e., the associated query) alongside the key itself.

The reason I made it an dumb array - for now - was for minimising the "need to know."  In my view it's only the scheme-specific logics (i.e., the plugin) that knows which keys it'll need to use, and therefore has to synthesise.  Whilst what the "core" VTS part that populates the kvstore needs to know is only that the supplied "val" will be aliased under the associated 1..n different "key"s.

common/scheme.go
setrofim: Not because the verifier doesn't interact with the scheme directly, but rather calls the `ITrustedServicesClient.Getattestation(*AttestationToken)` , the servier-side implementation for which then invokes the `IScheme.GetAttestation(*EvidenceContext, []string)`.

common/scheme_plugin.go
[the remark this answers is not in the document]
setrofim: Sure, and with a map you're still have the need-to-know as only the scheme knows how to interpret they keys, and from the out side code, it's basically an opaque `interface{}` either way. The textual, rather than numerical indexes just make make it easier for debugging.

Anyway, no biggie; just a thought. An array is totally fine for now.

common/scheme_plugin.go
```diff
@@ -74,13 +89,54 @@ func (s *SchemeServer) GetFormat(args interface{}, resp *AttestationFormat) erro
 	return nil
 }
 
+type SynthKeysArgs struct {
```
yogeshbdeshpande, line 92: @thomas-fossati I have opened a generic Issue #136 to address the potential concern I have regarding the way these arguments are handled in scheme_plugin.go

common/scheme_plugin.go
```diff
@@ -148,32 +206,76 @@ func (s *SchemeRPC) GetFormat() AttestationFormat {
...
+	return resp, nil
 }
 
 func (s *SchemeRPC) ExtractEvidence(token *AttestationToken, trustAnchor string) (*ExtractedEvidence, error) {
-	var err error
-	var args ExtractEvidenceArgs
+	var (
+		err  error
+		args ExtractEvidenceArgs
+	)
```
yogeshbdeshpande, line 278: Code at line 282, 290 and 297 is inconsistent with the way it has been done on 248 and 253. Could you please align both of these properly?

kvstore/memory.go
```diff
@@ -67,13 +67,14 @@ func (o *Memory) Set(key string, val string) error {
 	lk.Lock()
 	defer lk.Unlock()
 
-	_, ok := o.Data[key]
+	data, ok := o.Data[key]
 	if ok {
-		return fmt.Errorf("key %q already exist, delete it first", key)
+		// TODO(tho) -- check if the value is already stored, if so do not append
```
yogeshbdeshpande, line 72: Isn't the comment the opposite of what you are actually doing...?

common/scheme_plugin.go
thomas-fossati: OK.  Done in 73da0276

kvstore/memory.go
thomas-fossati: I think the comment is correct.  We are not checking if the value is already in the array at the moment, which is what the comment says.

kvstore/kvstore_test.go
```diff
@@ -0,0 +1,84 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package kvstore
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestKVStore_New_nil_config(t *testing.T) {
```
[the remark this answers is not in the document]
thomas-fossati: I guess it's a matter of style.  Maybe we could refactor at a later stage?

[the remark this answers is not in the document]
yogeshbdeshpande: Sure, please raise a github issue to refactor so that we do not lose it. Anyone touching file in future, can take the task and close the issue!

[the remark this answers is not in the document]
thomas-fossati: I am personally OK with the current style so I don't feel like an Issue is needed :-) But if you feel strongly that this needs attention please raise it.

kvstore/memory_test.go
```diff
@@ -208,8 +229,8 @@ func TestMemory_dump_ok(t *testing.T) {
 
 	expectedTbl := `Key                              Val
 ---                              ---
-psa://tenant-1/deadbeef/beefdead {"some": "json"}
-psa://tenant-2/cafecafe/cafecafe [1, 2, 3]
+psa://tenant-1/deadbeef/beefdead [{"some": "json"}]
```
yogeshbdeshpande, line 232: I thought this has been changed to psa-iot?

thomas-fossati: You can safely consider it just a random string with a vague scent of PSA :-)

common/scheme_plugin.go
thomas-fossati: Done again in [a4a676f](https://github.com/veraison/veraison/pull/134/commits/a4a676f99a4772e35dc82083f527d642b09df89a)

yogeshbdeshpande (approved):

kvstore/memory_test.go
thomas-fossati: From the point of view of the kvstore consistency with the PSA URL schema is completely irrelevant.

plugins/scheme-dice/main.go
```diff
@@ -32,6 +32,14 @@ func (s Scheme) GetTrustAnchorID(token *common.AttestationToken) (string, error)
 	return "dice://", nil
 }
 
+func (s Scheme) SynthKeysFromSwComponent(tenantID string, swComp *common.SwComponent) ([]string, error) {
```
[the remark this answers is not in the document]
thomas-fossati: Good point, this needs to be linked to a DICE "story", which at the moment we don't have.

setrofim (approved):

PR #138, opened 2022-03-30
fixup some unintended merges

SimonFrost-Arm:
Due to some historic oddity or finger trouble, I accidentally committed some ancient doc files in the last commit (and failed to merge one minor terminology change). This should fix that mess

Signed-off-by: Simon Frost <Simon.Frost@arm.com>


thomas-fossati (approved):
second time's a charm!

PR #148, opened 2022-04-06
Store API: remove distinction between id and value in endorsements

### LINKED ISSUE #122: Store API: remove distinction between id and value in the  endorsements

thomas-fossati:
* Remove distinction between id and value in endorsements
* Abstract SwComponents and TrustAnchors into a more regular Endorsement type
* Remove (for now) the unused Get{SwComponent,TrustAnchor} provisioning interfaces
* Also fix a couple of lint warnings in the provisioning area
* Remove IEvidenceExtractor interface, which is now OBE
* Remove GetEndorsements{Request,Response} as well as EndorsementID, which are OBE

Fixes #122


thomas-fossati:
@setrofim the updated internal interfaces look like this:
* [PSA](https://github.com/veraison/veraison/blob/id-val-merge/provisioning/plugins/corim-psa-decoder/README.md)
* [TPM-EnactTrust](https://github.com/veraison/veraison/blob/id-val-merge/provisioning/plugins/corim-tpm-enacttrust-decoder/README.md)

common/vts.proto
```diff
@@ -11,11 +11,9 @@ import "attestation.proto";
 import "token.proto";
 
 
-enum TAType {
```
yogeshbdeshpande, line 14: As you are tidying up vts.proto, I realized the following can also be tidied up quite easily.

Line: 21: There is no EndorsementID

Line 30: GetEndorsementsRequest() is NOT used any more [ I confirmed this with @setrofim : Thanks]

Line 40: GetEndorsementsResponse() is NOT used either (on the interface)

thomas-fossati: I think we should keep this PR scoped to what it says.

yogeshbdeshpande: Now that you are using the term Endorsement in your changes, this becomes extremely cluttered and confusing!

thomas-fossati: Anyway, I'll do this :)

yogeshbdeshpande: Anyway, I will do this post the merge of this one: See Issue #150 to handle this!

thomas-fossati: No need, it's already done in [31e8daa](https://github.com/veraison/veraison/pull/148/commits/31e8daae294514b9cea532bb14f0786e50f0e9d8)

common/vts.proto
```diff
@@ -11,109 +11,52 @@ import "attestation.proto";
...
-
-message GetSwComponentResponse {
-    Status status = 1;
-    google.protobuf.Struct attributes = 2;
-}
-
-message TrustAnchorID {
-    AttestationFormat type = 1;
-    google.protobuf.Struct parts = 2;
+enum EndorsementType {
```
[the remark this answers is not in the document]
yogeshbdeshpande: Otherwise, it gets very mis-aligned and confusing to the common reader of the interface!

common/vts.proto
```diff
@@ -11,109 +11,52 @@ import "attestation.proto";
...
 
-message TrustAnchorValue {
-    TAType type = 1;
-    google.protobuf.Struct value = 2;
-}
-
-message TrustAnchor {
-    TrustAnchorID id = 1;
-    TrustAnchorValue value = 2;
+message Endorsement {
```
yogeshbdeshpande, line 37: Also I am little nervous about the term Endorsement! Strictly speaking 
as per RATS Architecture, Endorsements are the entities which does not come in Evidence, but are elements used in Verification. 
Are we muddling up that distinction here ?

common/vts.proto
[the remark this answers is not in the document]
thomas-fossati: The idea was to align this with CoRIM

common/vts.proto
```diff
@@ -11,109 +11,52 @@ import "attestation.proto";
...
-
 // Client interface for the Veraison Trusted Services component.
 service VTS {
     // Returns attestation information -- evidences, endorsed claims, trust
     // vector, etc -- for the provided attestation token data.
     rpc GetAttestation(AttestationToken) returns (Attestation);
 
-    // Service endpoints that are used to store and retrieve Software Components
-    // and Trust Anchors to and from the endorsement store
+    // Service endpoints that are used to store Software Components
```
yogeshbdeshpande, line 57: If Trust Anchors and SW Component elements are in itself Endorsement as per your proposal, what is the benefit of retaining these two messages ?

I can just have one message:

AddEndorsements(EndrosementRequest) 
==> Type is enough to qualify what EndorsementType is been added ???

common/vts.proto
thomas-fossati: There is no established term that we can use to group endorsed values, reference values, verification and identity keys.  I have picked Endorsement as it's the term used in the DICE "Endorsements" architecture.

common/vts.proto
yogeshbdeshpande: The current store buckets are Endorsement Store and TrustAnchor store, so how this is aligned ?

thomas-fossati: The contract with the store layer is that SwComponents go to the Endorsements store while any crypto material goes into the TrustAnchor store.

thomas-fossati: It's tricky.  Make a naming proposal maybe?  And we can resolve the naming issue in a different PR.

yogeshbdeshpande: We need to be consistent to call out the store names to be aligned with the interface! 

Better to remove the name Endorsement Store and call it REFVAL Store 
Then it becomes neater and consistent!

Otherwise it is a mess ! We call both RefVal and TA as Endorsement but does not keep TA inside Endorsement Store ???

common/vts.proto
thomas-fossati: Also, we call the external interface [Endorsement Provisioning API](https://github.com/veraison/docs/blob/main/api/endorsement-provisioning/README.md).

common/vts.proto
thomas-fossati: > Better to remove the name Endorsement Store and call it REFVAL Store Then it becomes neater and consistent!

I am not sure.  I think the same store could end up containing "endorsed values" (e.g., security certification status).

> Otherwise it is a mess ! We call both RefVal and TA as Endorsement but does not keep TA inside Endorsement Store ???

As I said, naming this stuff is a tricky thing, mainly because the existing terminology is quite lacking.

common/vts.proto
```diff
@@ -11,109 +11,52 @@ import "attestation.proto";
...
+    // and Trust Anchors to the endorsement store
```
yogeshbdeshpande, line 58: Thinking a bit more, as this is a generic interface and not tied to any specific attestation scheme:
Ideally this should not be named as SW Components, but AddReferenceVals()

Reason been, in a particular attestation scheme, the attestation may be other types of reference values which we want to provision but they may not be SW Components!

common/vts.proto
thomas-fossati: Veraison needs to be able to distinguish cryptographic endorsements (which may need a different treatment in terms of storage security guarantees) from the rest.

common/vts.proto
thomas-fossati: Yeah, that may be a slightly better name.

common/vts.proto
yogeshbdeshpande: > > Better to remove the name Endorsement Store and call it REFVAL Store Then it becomes neater and consistent!
> 
> I am not sure. I think the same store could end up containing "endorsed values" (e.g., security certification status) alongside "reference values".
>> Why? We could have endorsed-vals store ?
> 
> > Otherwise it is a mess ! We call both RefVal and TA as Endorsement but does not keep TA inside Endorsement Store ???
> 
> As I said, naming this stuff is a tricky thing, mainly because the existing terminology is quite lacking.

common/vts.proto
thomas-fossati: Anyway, I suggest we have a separate "naming consistency" conversation (where we may also want to discuss things like using "Evidence" instead of "Attestation Token" in the verification pipeline) and unblock this PR.

common/vts.proto
thomas-fossati: how many stores do we want to maintain? :-)

thomas-fossati: It is a serious question though: I think having a "cryptographic endorsements" and "normal endorsements" should be enough.

common/vts.proto
yogeshbdeshpande: how big is an impact just of changing SwComponent in the interface to RefVal ? I agree for other bits, we need a different PR, but can we consider, the impact before concluding this change also (via separate PR?)

thomas-fossati: The impact is a search and replace on ~20 files.

setrofim: > The distinction is important in the sense that they are supposed to go into different store buckets. The plugin knows where they need to be routed.

This can be done based on the "type" field of the Endorsement,  and does not necessarily require a separate method. Using a single message would actually be more flexible, as there this means we don't actually have to define the number of stores architecturally -- this can be left to the implementation (e.g. what is the issue with keeping trust anchors and ensorsements/refvals in the same store?)

yogeshbdeshpande: For addressing this specific concern I have raised Issue #151. However , in the vts.proto SwComponent to RefVal change is quite a logical change which can go first with this change, unless you feel need a more careful deliberation?

setrofim (approved):

common/vts.proto
yogeshbdeshpande: Please refer Issue #152 for further Architectural Discussion

yogeshbdeshpande (approved):

PR #166, opened 2022-07-17
verifier REST API

thomas-fossati:
This commit fully implements the challenge-response REST API defined in [1].

[1] https://github.com/veraison/docs/tree/main/api/challenge-response

Signed-off-by: Thomas Fossati <thomas.fossati@arm.com>


verification/api/challengeresponsesession.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err := json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+
+	return o.FromString(s)
+}
+
+type Blob struct {
+	Type  string `json:"type"`
```
yogeshbdeshpande, line 65: I suppose it is media type? if yes can it be named as mtype for json  type?

verification/api/challengeresponsesession.go
```diff
@@ -0,0 +1,92 @@
...
+	Value []byte `json:"value"`
+}
+
+type ChallengeResponseSession struct {
+	id            string
+	Status        Status    `json:"status"`
```
yogeshbdeshpande, line 71: On line 71 just a question: If Status is the element we control and inside the structure, is there a real need for it to be having an Integer ? Can it not be a set of acceptable strings which we define and rest populate as an error. This way, we avoid the conversion on line 20 and line 34

verification/api/challengeresponsesession.go
```diff
@@ -0,0 +1,92 @@
...
+	Nonce         []byte    `json:"nonce"`
+	Expiry        time.Time `json:"expiry"`
+	Accept        []string  `json:"accept"`
+	Evidence      *Blob     `json:"evidence,omitempty"`
+	Result        *[]byte   `json:"result,omitempty"`
```
yogeshbdeshpande, line 76: Could line 76: Result be named as AttestationResult, Otherwise it is very confusing to have 
(a) Status (b) Result (c) Failure Reason ?

verification/api/challengeresponsesession.go
```diff
@@ -0,0 +1,92 @@
...
+	FailureReason *string   `json:"failure-reason,omitempty"`
```
yogeshbdeshpande, line 77: Can the FailureReason be clubbed with Status in a more meaningful way rather than hanging here in isolation. Unless I am missing something, this is the reason when Status is "Failed" and you want to elaborate more on the Reason why we failed?

verification/api/challengeresponsesession.go
yogeshbdeshpande: And please explain with comment that it is Media Type

verification/api/handler.go
```diff
@@ -0,0 +1,385 @@
...
+
+	uuid.SetNodeID([]byte(mid))
+
+	return uuid.NewUUID()
+}
+
+// mintNonce creates a random nonce of nonceSz bytes.  nonceSz must be strictly
+// positive
+func mintNonce(nonceSz uint8) ([]byte, error) {
+	if nonceSz == 0 {
```
yogeshbdeshpande, line 77: We should check here for acceptable nonceSz right?

verification/verifier/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/veraison/services/verification/verifier
+
+go 1.18
```
yogeshbdeshpande, line 3: Question: Are we sticking with 1.17 or go 1.18?

verification/sessionmanager/sessionmanager_ttlcache.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2022 Contributors to the Veraison project.
```
yogeshbdeshpande, line 1: Just a question on file naming.

As the file is in sessionmanager, can we change the file to 

just ttlcache.go

verification/api/challengeresponsesession.go
thomas-fossati: This is the `EvidenceBlob` described in the OpenAPI doc: https://github.com/veraison/docs/blob/main/api/challenge-response/schemas/components.yaml

verification/api/challengeresponsesession.go
thomas-fossati: we'd loose the strong typing feature we currently have

verification/api/challengeresponsesession.go
thomas-fossati: this also is a 1:1 map of the OpenAPI description: https://github.com/veraison/docs/blob/main/api/challenge-response/schemas/components.yaml

verification/api/challengeresponsesession.go
thomas-fossati: I don't know why I added this, it's not in the API.  We should drop it.

verification/api/challengeresponsesession.go
```diff
@@ -0,0 +1,92 @@
...
+	if err := json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+
+	return o.FromString(s)
+}
+
+type Blob struct {
+	Type  string `json:"type"`
+	Value []byte `json:"value"`
```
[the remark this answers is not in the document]
thomas-fossati: It's the API doc that dictates here and everywhere else we are dealing with the external interface.  I'll add an explicit link to the API documentation.

verification/api/handler.go
thomas-fossati: I don't understand this comment, sorry.

verification/verifier/go.mod
thomas-fossati: we want 1.18 for the generics on which the TTL cache depends

verification/verifier/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/veraison/services/verification/verifier
```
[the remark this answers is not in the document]
thomas-fossati: It doesn't matter at this level, because the go.mod will handle the re-routing.

verification/api/challengeresponsesession.go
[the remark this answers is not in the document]
yogeshbdeshpande: Ok, that might help to some extent!

verification/api/handler.go
yogeshbdeshpande: I mean Veraison does not accept any NonceSz but only few right, 32, 64 and 128

thomas-fossati: oh, where did we write that?

thomas-fossati: that'd be utterly wrong

verification/api/challengeresponsesession.go
thomas-fossati: done in 2f1832a

verification/api/handler.go
yogeshbdeshpande: description:
            base64 encoded 32, 48, or 64 bytes random value

verification/verifier/go.mod
yogeshbdeshpande: Does the CI works with new go revision?

verification/api/challengeresponsesession.go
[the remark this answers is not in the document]
thomas-fossati: also added the spec ref in fecc96f

verification/verifier/go.mod
thomas-fossati: I'd hope so, why?

yogeshbdeshpande: We had an issue in CI tests when we moved to 1.18 : Remember the conversation we had about this..

verification/api/challengeresponsesession.go
yogeshbdeshpande: Correct: All I am saying that change Result in go field to Attestation Result:
========================

          $ref: '#/components/schemas/EvidenceBlob'
          description: >
            the submitted attestation Evidence
        result:
          $ref: '#/components/schemas/AttestationResult'

============================
Also where is the AttestationResult enumerated?  As per the API documentation?

verification/verifier/go.mod
thomas-fossati: not really, my memory is in terrible state :(

verification/api/challengeresponsesession.go
thomas-fossati: > Correct: All I am saying that change Result in go field to Attestation Result:

it's called `result` in the API, why should we deviate?

> Also where is the AttestationResult enumerated? As per the API documentation?

the format of the result is determined by the Verification module, the API frontend will forward whatever it's given.

yogeshbdeshpande (approved):
LGTM!

